PR #3, opened 2015-11-25
Add Python 3.5 to the testing matrix and also fix tests for PyQt5

ccordoba12:


ccordoba12:
I'm a bit surprised tests were deactivated here, but now they on!

PR #8, opened 2016-01-14
QtGui: Unify it with QtWidgets for PyQt5

ccordoba12:
- This is how it's done in qtconsole and Spyder, so I'd like to follow this convention here too.
- This is causing me a lot of headaches when developing external projects with `qtpy`. So either QtWidgets is removed (which I'd prefer) or QtGui has to go, but we must not expose both.


ccordoba12:
Sorry, now I understand how `qtpy` works: everything should be developed as if we were working with PyQt5.

So this PR is really not necessary.

PR #9, opened 2016-01-14
Add improvements and missing imports from Spyder Qt shim

### LINKED ISSUE #5: Add PYQT5 / PYQT4/PYSIDE boolean variables
Spyder has a number of global variables for convenience to check whether we are running on PyQt5, PyQt4 or PySide.
It would be nice if those were moved to qtpy. I think that anyone using qtpy could use those.

ccordoba12:
Fixes #5

----

Spyder is much more battle tested than qtpy or pyqode, so I prefer to use as much as we can from its shim here.

### DIFF
--- a/qtpy/QtCore.py
+++ b/qtpy/QtCore.py
@@ -11,6 +11,7 @@ Provides QtCore classes and functions.
 """
 
 import os
+
 from qtpy import QT_API
 from qtpy import PYQT5_API
 from qtpy import PYQT4_API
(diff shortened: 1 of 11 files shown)


ccordoba12:
The main difference now is that `QT_API` has to be in lower caswe, but that's the custom elsewhere.

ccordoba12:
Another important thing is that I replaced most of the current `qtpy/__init__.py` with `spyderlib/qt/__init__.py` because the first one was not working and second is much more stable.

ccordoba12:
Finally, given the importance of this change, I think we should bump `qtpy` version to **0.2.0**

ccordoba12:
I'd even say that we should bump `qtpy` version to **1.0** because after these changes and those in #10 it should be pretty stable :-)

ccordoba12:
Of course, my main interest with all this is to start using `qtpy` in Spyder itself since beta3 ;-)

goanpeca:
LGTM :+1:

PR #13, opened 2016-02-13
Add support for QtWebEngine widgets

ccordoba12:
Those widgets are based on Chromium and were introduced in Qt 5.4


ccordoba12:
Our current builds of Qt5 doesn't include those widgets, but they are certainly the future of Web widgets in Qt.

Issue #21, opened 2016-03-30
Patch QComboBox with PySide?


astrofrog:
In qt-helpers, we patch QComboBox because on PySide, storing Python objects in the ``userData`` causes segmentation faults (while it works correctly in PyQt4 and PyQt5). Here is our implementation (which is only used with PySide):

https://github.com/glue-viz/qt-helpers/blob/master/qt_helpers.py#L320

Would you be interested in something similar here? If so, I can open a PR.

ccordoba12:
We were not aware of that issue, but if you recognized it as such in Glue then I think it's important to have it in `qtpy` too.

So, yes, please open a PR for it :-)

Nodd:
Just to be sure, did you check that the bug still exists in the last version ?
If that's the case a PR is welcomed !

astrofrog:
Ok, I'll work on it at the weekend! (and I'll check the bug is still present)

goanpeca:
Hi @astrofrog just checking if you found the issue :-p

astrofrog:
@goanpeca - oops, haven't had time yet, will try and look shortly.

astrofrog:
I'm finally working on switching glue to use QtPy, and will report back on this soon.

ccordoba12:
Nice!! Please send us as many PR as you consider necessary to support 
Gue :-)

PR #25, opened 2016-06-08
Add patched version of QComboBox

astrofrog:
In PySide, using Python objects as userData in QComboBox causes Segmentation faults under certain conditions. Even in cases where it doesn't, findData does not work correctly. Likewise, findData also does not work correctly with Python objects when using PyQt4. On the other hand, PyQt5 deals with this case correctly. We therefore patch QComboBox when using PyQt4 and PySide to avoid issues.

I'm starting off by adding only the regression test to make sure that it fails as expected.

Closes https://github.com/spyder-ide/qtpy/issues/21


astrofrog:
Interestingly, the regression test is segfaulting for PyQt4 and PyQt5 too here, which is weird, since I didn't see the segfault on PyQt5 before.

astrofrog:
Ah, I think I know what might be causing that - will try and fix it tomorrow.

astrofrog:
Ok, so the regression test now behaves as expected (I had forgotten to start up QApplication):

- It causes a segfault with PySide
- It gives the wrong result with PyQt4
- It works with PyQt5

astrofrog:
@ccordoba12 - I'm not sure what the right place is to add these kinds of patches - currently I just put it inside ``QtWidgets.py`` directly, but let me know if you would like it to be kept elsewhere.

astrofrog:
@ccordoba12 - as you can see, Travis now passes. I don't think we'll be able to pass all the quantifiedcode tests, since some of the function names, to match Qt, are un-Pythonic. In any case, this is ready for review!

Nodd:
QuantifiedCode is informative. You can safely ignore the warnings.

astrofrog:
One thing I did add here was that Travis now actually runs ``py.test`` in the ``tests`` directory. I think we could also make the existing ``tests.py`` file into an actual tests, i.e. checking the string representation of the classes based on the environment variables. I could do that in a separate PR?

astrofrog:
Since I'm not the sole copyright holder, I've put a copy of the qt-helpers license in the function, but I'm happy to do it differently if there is a better way?

Nodd:
Yes, it would be nice to have real tests using pytest. PRs are always welcomed !

goanpeca:
@astrofrog thanks for this :-), as @nodd says quantified code is informative, and the non pythonic names thing will be solved when they include some sort of exclusion mechanism, or we could just disable that rule for good :-p.

Py Tests always welcome :-)

qtpy/QtWidgets.py
```diff
@@ -22,6 +22,110 @@
 from qtpy import PythonQtError
 
 
+def patch_qcombobox():
+    """
```
ccordoba12: I'd prefer this function to leave in a different file. It could be called `pyside_combobox` or something like that :-)

@Nodd, @goanpeca, what do you think?

astrofrog: That would be fine by me, but I'll wait until others confirm. Maybe the filename could start with ``patch_``, and then other similar patches would use the same convention?

Nodd: Since the function has to be on top it makes the module layout a bit strange. Maybe better than `patch*` would be a submodule `_patch/` so that it doesn't pollute the main directory.

ccordoba12: +1 to @Nodd's suggestion.

astrofrog:
@Nodd @ccordoba12 - I've now moved the patch to ``qtpy._patch.qmessagebox``

ccordoba12:
I think it's fine now, but you need to rebase or merge with master to solve conflicts.

astrofrog:
@ccordoba12 - done!

ccordoba12:
Thanks!

PR #28, opened 2016-06-09
Make tests.py into proper unit test, and add Qt version info to pytest header

astrofrog:
Since I'm adding actual tests in #25 and #27, I thought it would make sense to formally switch over to using py.test to run tests. This also makes it so that the py.test header includes detailed Qt version information:

```
============================= test session starts ==============================
platform darwin -- Python 3.5.1, pytest-2.8.5, py-1.4.31, pluggy-0.3.1
benchmark: 3.0.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=5.00us max_time=1.00s calibration_precision=10 warmup=False warmup_iterations=100000)

PyQt4: PyQt: 4.11.4 - Qt: 4.8.7
PyQt5: not installed
PySide: not installed
...
```


astrofrog:
This is now ready for review! (I think the quantifiedcode warnings can be ignored in this case)

goanpeca:
Great work @astrofrog, :-)

ccordoba12:
@astrofrog, thanks a lot for working on this! It's a great improvement :-)

@goanpeca, please disable QuantifiedCode.

tests/test_main.py
```diff
@@ -0,0 +1,57 @@
...
+
+    QT_API = os.environ.get('QT_API', None)
+
+    if QT_API == 'pyside':
+        assert_pyside()
+    elif QT_API in ('pyqt', 'pyqt4'):
+        assert_pyqt4()
+    elif QT_API == 'pyqt5':
+        assert_pyqt5()
+    else:
```
ccordoba12: Is it possible to enter this part if we set always set `USE_QT_API` in `travis.yml`?

astrofrog: No, but it's possible if one runs the tests locally without specifying USE_QT_API or QT_API

ccordoba12: Ok

astrofrog:
@ccordoba12 - do you need me to squash the commits? (not sure what kind of policy you use for qtpy compared to other projects)

goanpeca:
@ccordoba12 the quantified code suggestions are valid 

- Consider documenting your function(s)
- Avoid `None` as a redundant second argument to `dict.get()`

I will disable the one that complaints about having non pythonic method names which does not work when using Qt.

@astrofrog could you make the suggested "fixes"?

ccordoba12:
Ok, I think this one is ready. We have to merge it with AppVeyor failing because there's no AppVeyor working at this point :-)

PR #30, opened 2016-06-10
Added is_py* helper functions to help determine which Qt bindings are in use

astrofrog:
This is pretty simple, and you could argue maybe they aren't needed - but for me they are quite useful because it's easier to remember ``is_pyside`` rather than figure out what the internal API variables are :) (and using ``is_pyside`` in code looks cleaner). There are some rare cases where even though we're using qtpy, we need to know which binding we are using.

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -127,3 +127,27 @@ if API in PYSIDE_API:
         PYSIDE = True
     except ImportError:
         raise PythonQtError('No Qt bindings could be found')
+
+
+def is_pyside():
+    """
+    Helper function that returns a boolean indicating whether we are using
+    PySide.
+    """
+    return API in PYSIDE_API
+
+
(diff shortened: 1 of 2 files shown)


goanpeca:
yes I like this :-)

I also wanted some time ago to have, is_win, is_mac, is_unix, is_linux, but @ccordoba12 did not agree :-p

ccordoba12:
@astrofrog, this is already implemented but using constants instead of functions. Look in the same file for `PYQT5`, `PYQT4` and `PYSIDE` :-)

Since we already use these constants in Spyder, I'm -1 for this PR.

astrofrog:
@ccordoba12 - that works for me, feel free to close this

Nodd:
Yeah, constants already exist, I'm closing it. Thanks for taking the time to make the PR though !

PR #33, opened 2016-06-10
Ignore case for QT_API env variable in qtpy submodules

Nodd:
This is easier on the users, and the tests are currently failing because of this problem.

I got rid of the `os.environ[QT_API]` lookups, because it makes qtpy rely too much on an external global state, while the environent is already pared in `__init__.py`. This makes the import thread safe, for example.

With this I think that https://github.com/spyder-ide/spyder/pull/3210 won't fail anymore.

### DIFF
--- a/qtpy/QtCore.py
+++ b/qtpy/QtCore.py
@@ -10,20 +10,19 @@
 Provides QtCore classes and functions.
 """
 
-from qtpy import API
-from qtpy import PYQT5_API
-from qtpy import PYQT4_API
-from qtpy import PYSIDE_API
+from qtpy import PYQT5
+from qtpy import PYQT4
+from qtpy import PYSIDE
 from qtpy import PythonQtError
 
 
-if API in PYQT5_API:
+if PYQT5:
     from PyQt5.QtCore import *
     from PyQt5.QtCore import pyqtSignal as Signal
     from PyQt5.QtCore import pyqtSlot as Slot
     from PyQt5.QtCore import pyqtProperty as Property
     from PyQt5.QtCore import QT_VERSION_STR as __version__
-elif API in PYQT4_API:
+elif PYQT4:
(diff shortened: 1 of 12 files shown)


qtpy/QtCore.py
```diff
@@ -10,22 +10,20 @@
...
 """
 
-import os
-
-from qtpy import QT_API
+from qtpy import API
 from qtpy import PYQT5_API
 from qtpy import PYQT4_API
 from qtpy import PYSIDE_API
 from qtpy import PythonQtError
```
Nodd: BTW, why are those imports split into multiple lines ?

qtpy/QtCore.py
```diff
@@ -10,22 +10,20 @@
...
 
 
-if os.environ[QT_API] in PYQT5_API:
+if API in PYQT5_API:
```
astrofrog: Why not use just the ``PYQT5`` variable here?

Nodd: You're right, that's even better. I was too focused on `os.environ`.

qtpy/QtCore.py
```diff
@@ -10,22 +10,19 @@
...
 
-import os
-
-from qtpy import QT_API
-from qtpy import PYQT5_API
-from qtpy import PYQT4_API
-from qtpy import PYSIDE_API
+from qtpy import PYQT5
+from qtpy import PYQT4
+from qtpy import PYSIDE
```
astrofrog: I agree with your previous comment that this could maybe be combined into a single line

ccordoba12: Yep, I agree with @astrofrog

ccordoba12:
I like this one. Please unify `PYQT5`, etc imports in a single line and fix conflicts to merge it :-)

Nodd:
It should be good, waiting for tests.

PR #37, opened 2016-06-12
Automatically load custom widget classes when using PySide

astrofrog:
This is a follow-up to https://github.com/spyder-ide/qtpy/pull/27 (and will be rebased once that is merged) which includes auto-loading of custom widget classes by parsing the ui files. This then makes the behavior of loadUi consistent with that in PyQt4/5.

This is still a work in progress - I need to test this extensively with glue to make sure that it always works. I'm also not entirely happy with the sys.path hack, and want to try and find a better solution. As such, this is not ready for a proper review, but any preliminary feedback/suggestions would be useful (will be easier to see the changes here once #27 is merged)

### DIFF
--- a/qtpy/uic.py
+++ b/qtpy/uic.py
@@ -171,10 +171,10 @@ elif PYSIDE:
         # Get the customwidgets section
         custom_widgets = ui.find('customwidgets')
 
-        custom_widget_classes = dict()
-
         if custom_widgets is None:
-            return custom_widget_classes
+            return {}
+
+        custom_widget_classes = {}
 
         for custom_widget in custom_widgets.getchildren():
 
--- a/qtpy/uic.py
+++ b/qtpy/uic.py
@@ -154,8 +154,40 @@ elif PYSIDE:
 
                 return widget
 
-    def loadUi(uifile, baseinstance=None, customWidgets=None,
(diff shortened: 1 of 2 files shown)


Nodd:
You can merge/rebase this one !

astrofrog:
@Nodd - this is now rebased and ready for review!

qtpy/uic.py
```diff
@@ -154,8 +154,40 @@ def createWidget(self, class_name, parent=None, name=''):
...
+        etree = ElementTree()
+        ui = etree.parse(ui_file)
+
+        # Get the customwidgets section
+        custom_widgets = ui.find('customwidgets')
+
+        custom_widget_classes = dict()
+
+        if custom_widgets is None:
+            return custom_widget_classes
```
Nodd: Suggestion for clarification:
```python
        if not custom_widgets:
            return {}

        custom_widget_classes = {}
```
It's clearer that an empty dict is returned.

astrofrog:
@Nodd - I addressed your comment, so this is ready for a final review/merging

Nodd:
The test is passing, I merge. If there are corner cases, they will be discovered after some use.
With this I'll switch to qtpy for some of my projects, thanks a lot !

PR #41, opened 2016-06-15
qtpy.__version__ should be QtPy version, not Qt version

astrofrog:
I think that ``__version__`` should be set to the actual QtPy version, otherwise packages won't be able to easily check for example that QtPy is recent enough to include certain patches. I think we could then make the Qt and other versions available via e.g.

```
qtpy.QT_VERSION
qtpy.PYSIDE_VERSION
qtpy.PYQT4_VERSION
qtpy.PYQT5_VERSION
```

I'd be happy to include a quick PR to include in QtPy 1.1 if others agree?

Of course, we could have ``qtpy.QTPY_VERSION`` be the QtPy version, but I think it would be counter-intuitive for ``__version__`` to not be that.

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -52,6 +52,9 @@ packages::
 
 import os
 
+# Version of QtPy
+__version__ = "1.1.dev0"
+
 #: Qt API environment variable name
 QT_API = 'QT_API'
 #: names of the expected PyQt5 api
@@ -79,7 +82,8 @@ class PythonQtError(Exception):
 
 if API in PYQT5_API:
     try:
-        from PyQt5.QtCore import PYQT_VERSION_STR as __version__
+        from PyQt5.Qt import PYQT_VERSION_STR as PYQT_VERSION  # analysis:ignore
+        from PyQt5.Qt import QT_VERSION_STR as QT_VERSION  # analysis:ignore
     except ImportError:
         API = 'pyqt'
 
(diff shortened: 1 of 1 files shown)


astrofrog:
Code attached - note that for now, I actually went with:

```
qtpy.QT_VERSION
qtpy.PYSIDE_VERSION
qtpy.PYQT_VERSION
```

I could easily split ``PYQT_VERSION`` into ``PYQT4_VERSION`` and ``PYQT5_VERSION``, but do we want that? (if so, do we want ``QT4_VERSION`` and ``QT5_VERSION``?)

astrofrog:
I'm not sure what convention you want to use for the ``__version__`` number, but I went with what I would use for other projects - let me know if you want to change that.

Nodd:
What did you do to transform the initial issue into a PR ? I didn't know that it was possible.

astrofrog:
✨ Magic ✨ 

Actually I just used the ``hub`` command :smile:

astrofrog:
Once this is merged, then from my point of view we're good for a 1.1 release from the point of view of glue (all our tests pass).

Nodd:
Only one of `PYSIDE_VERSION` and `PYQT_VERSION` are defined at a time. I think it's better if the values are set to `None` by default rather than not being defined. Getting `None` is more understandable than getting an `AttributeError`.

astrofrog:
@Nodd - oops, of course, sorry about that

qtpy/__init__.py
```diff
@@ -52,6 +52,9 @@
 
 import os
 
+# Version of QtPy
+__version__ = "1.1.dev0"
```
ccordoba12: `__version__` is defined here:

https://github.com/spyder-ide/qtpy/blob/master/qtpy/_version.py#L2

Please use that here instead :-)

ccordoba12:
Just one minor comment, after that we can merge and release 1.1

qtpy/__init__.py
astrofrog: Done!

astrofrog:
@ccordoba12 - this should be ready to go

Nodd:
@ccordoba12 I merged, you can do the release !

ccordoba12:
I think I need to do first a PR in Spyder because of this change. I'll let you know about it :-)

PR #43, opened 2016-06-20
Don't check PyQt version with qtpy's version for old PyQt versions

### LINKED ISSUE #42: Wrong old PyQt4 version check
Maybe I'm wrong, but it seems to me that `is_old_pyqt` and `is_pyqt46` should use `PYQT_VERSION` instead of `__version__` in https://github.com/spyder-ide/qtpy/blob/master/qtpy/__init__.py#L113.  `__version__` is now the version of qtpy.

Nodd:
Fixes #42. This change follows #41.

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -110,8 +110,8 @@ if API in PYQT4_API:
     except ImportError:
         API = 'pyside'
     else:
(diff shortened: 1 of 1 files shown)


Nodd:
BTW I did the change but I don't have these old pyqt versions to test.

ccordoba12:
Thanks @Nodd!

PR #44, opened 2016-07-01
Make qtpy to set the QT_API environment variable

ccordoba12:
- Since qtpy is just a wrapper, there's no other way to tell external processes started by a PyQt/PySide application the Qt toolkit its parent is running on.
- Because of this, not setting QT_API was breaking Spyder's IPython kernels and Python consoles.

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -67,7 +67,8 @@ PYQT4_API = [
 #: names of the expected PySide api
 PYSIDE_API = ['pyside']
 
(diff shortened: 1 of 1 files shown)


ccordoba12:
Perhaps *no other way* is too much to say. But it's easier and more convenient for qtpy to set `QT_API` than for applications to set it after looking for which of `PYQT5`, `PYQT4` or `PYSIDE` qtpy constants is `True` :-)

Nodd:
You forgot `.lower()` so that the env variable is case insensitive.

PR #45, opened 2016-07-04
Allow QT_API values that are not completely in lower case

ccordoba12:
@Nodd, I think this is enough, right?


Nodd:
Yes that's it !

PR #47, opened 2016-07-06
Remove PyQt variant symbols from QtCore

### LINKED ISSUE #35: del some names in QtCore ?
In `QtCore.py` a few variables are renamed but the old names still persist, like `pyqtSignal` and `Signal`. PyQt variants should be removed.
(PyQt variants are imported by `import *`)

### LINKED ISSUE #36: Simplify QtCore.py
There is some coe like:
```python
    from PyQt4.QtCore import *
    from PyQt4.QtCore import QCoreApplication
```
Why is the secon line needed, since `import *` is supposed to have imported `QCoreApplication` already ?

Nodd:
Fixes #36.
Fixes #35.

I still think that the duplicate imports are very weird and not needed, But it doesn't hurt too much to leave them.

### DIFF
--- a/qtpy/QtCore.py
+++ b/qtpy/QtCore.py
@@ -24,6 +24,8 @@ if PYQT5:
     del pyqtSignal, pyqtSlot, pyqtProperty, QT_VERSION_STR
 elif PYQT4:
     from PyQt4.QtCore import *
+    # Those are things we inherited from Spyder that fix crazy crashes under
+    # some specific situations. (See #34)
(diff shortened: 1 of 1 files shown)


ccordoba12:
Pining @goanpeca because he's trying to use `QT_VERSION_STR` in Navigator.

goanpeca:
Oh right I should have used th QT4 constants

Nodd:
Let's wait that the navigator is fixed before merging then.

goanpeca:
Fixed on my side

PR #55, opened 2016-08-03
PR: Add explicit imports to QtSvg module

goanpeca:
@ccordoba12, @Nodd should we delete `PYQT4, PYQT5, PYSIDE, PythonQtError` from the module at the end?

`del PYQT4, PYQT5, PYSIDE, PythonQtError` ?

### DIFF
--- a/qtpy/QtSvg.py
+++ b/qtpy/QtSvg.py
@@ -1,21 +1,23 @@
 # -*- coding: utf-8 -*-
-#
+# -----------------------------------------------------------------------------
 # Copyright © 2009- The Spyder Development Team
+#
 # Licensed under the terms of the MIT License
 # (see LICENSE.txt for details)
+# -----------------------------------------------------------------------------
+"""Provides QtSvg classes and functions."""
 
-"""
-Provides QtSvg classes and functions.
-"""
-
-from qtpy import PYQT5, PYQT4, PYSIDE, PythonQtError
-
+# Local imports
+from qtpy import PYQT4, PYQT5, PYSIDE, PythonQtError
 
 if PYQT5:
-    from PyQt5.QtSvg import *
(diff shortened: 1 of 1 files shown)


goanpeca:
@ccordoba12, @Nodd should we delete `PYQT4, PYQT5, PYSIDE, PythonQtError` from the module at the end?

`del PYQT4, PYQT5, PYSIDE, PythonQtError` ?

ccordoba12:
> should we delete PYQT4, PYQT5, PYSIDE, PythonQtError

I don't think so because they don't belong to the module itself. @Nodd?

goanpeca:
Ehh ??
 precisely because they dont belong and would remain in the namespace is that I am asking if we should delete them.

ccordoba12:
I really don't see the need of it. That's normally how Python works, i.e. imported symbols remain in the namespace.

goanpeca:
Yes, I know how it works... this not about how python works or does not, but a user would be able to import PYQT4 from QtCore module and it should not, cause those symbols live in qtpy. not int qtpy.QtCore.

ccordoba12:
I doubt anyone would confuse our constants with proper Qt objects, but I'm neutral about removing or leaving them.

Besides, this needs a rebase now ;-)

Nodd:
I think that deleting them is indeed cleaner. I don't hink that many users would import it from here, but for example `vars(QtSvg)` shouldn't show it.

ccordoba12:
@goanpeca, would you mind to rebase and finish this one so we can merge it?

I agree with @Nodd on removing the PY* constants on every `qtpy` module :-)

goanpeca:
ok , will do :-)

goanpeca:
@ccordoba12, see :-p, its the fastest by far

image

ccordoba12:
This looks good to me :-)

PR #56, opened 2016-08-09
PR: Undo breaking change

goanpeca:
the .ciocheck file is the linter that will be enabled soon, (I was testing it...) it is harmless for now

### DIFF
--- a/.ciocheck
+++ b/.ciocheck
@@ -1,0 +1,58 @@
+# -----------------------------------------------------------------------------
+# isort
+# https://github.com/timothycrosley/isort/wiki/isort-Settings
+# -----------------------------------------------------------------------------
+[settings]
+from_first = true
+import_heading_stdlib = Standard library imports
+import_heading_firstparty = Local imports
+import_heading_thirdparty = Third party imports
+indent = '    '
+known_third_party=six
+line_length = 79
+sections = FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
+
+# -----------------------------------------------------------------------------
+# Flake 8
+# http://flake8.readthedocs.io/en/latest/config.html
+# -----------------------------------------------------------------------------
+[flake8]
+exclude = */tests/*
+ignore = E126, F401
+max-line-length = 79
+max-complexity = 64
+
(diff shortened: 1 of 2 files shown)


goanpeca:
@ccordoba12 could we make a new release after this is merged to avoid the breakage?

ccordoba12:
I don't agree with this change. What @Nodd did in PR #47 was the right thing to do to make things work the same way in PyQt and PySide. Besides, only requiring this for PyQt4 and not PyQt5 is ugly too.

My point is that `qtpy` needs to be treated as an upstream project by all those that depend on it (and not as a downstream one, as you're trying to do here :-). This means that it's development is independent from what happens in Navigator, Spyder, Glue and git-cola, and we shouldn't make changes here just to satisfy one or other of these projects unless it's really justified by them (like my introduction of the `WEBENGINE` constant for Spyder, or the addition of the `ui` module for Glue).

Besides, you have several choices to fix this problem in Navigator:

1. Pin the latest version of `qtpy` that works with Navigator in the Navigator conda recipe (by bumping the build number of the current version).
2. Release a minor version of Navigator to fix this problem.
3. Ask Ilan to not upload a new version of `qtpy` to Continuum's default channels.

Fortunately, none of these options requires releasing a new version of `qtpy` that doesn't add a real contribution to its development :-)

ccordoba12:
And another thing: please stop making in-repo branches for your PRs. I'd really prefer that you work in your own fork and send PRs from there :-)

At the end that is as simple as adding a new `remote` to your git config ;-)

goanpeca:
>And another thing: please stop making in-repo branches for your PRs. I'd really prefer that you work in your own fork and send PRs from there :-)

No double builds were activated because the appveyor and travis recipe are constrained to only build master branches. So what is the now, really?

ccordoba12:
> So what is [problem] the now, really?

It's a just a different working style. As I said before, all open source projects I follow (IPython, pandas, mpl) use in-repo branches only to:

1. Differentiate between stable and unstable lines of development
2. Work on some hard to develop feature that requires a shared branch for the whole team.

So I'd prefer to follow this style for all our projects too.

goanpeca:
Mehhh, ok

PR #71, opened 2016-09-19
PR: Use relative imports so its vendored more easily

almarklein:
I only need QtCore, QtGui, QtWidgets, so it would be nice if I can vendor-in part of the qtpy project. Having relative imports makes this much easier.

### DIFF
--- a/qtpy/QtCore.py
+++ b/qtpy/QtCore.py
@@ -10,7 +10,7 @@
 Provides QtCore classes and functions.
 """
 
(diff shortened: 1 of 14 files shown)


goanpeca:
@ccordoba12 or @Nodd any comments?

Nodd:
Very good idea! And good ideas are even better with a PR :)

ccordoba12:
I agree with @Nodd. @Nodd, could you review @almarklein's work and merge it? I'm a bit busy right now :-)

Nodd:
Done ! Thanks !

PR #72, opened 2016-09-26
Include QStringListModel in QtCore

almarklein:


almarklein:
Maybe it should also be deleted from QtGui?

ccordoba12:
@almarklein, you also need to remove these objects from `qtpy.QtWidgets`

almarklein:
>  you also need to remove this object from qtpy.QtWidgets

Yeah, but cannot edit other files if I created this PR directly from GH :) I created #74.

PR #107, opened 2017-05-04
PR: Handle C style header for custom widget

diorcety:
Convert c style header in python import (like pyside-uic)

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -85,7 +85,7 @@ PYSIDE_API = ['pyside']
 PYSIDE2_API = ['pyside2']
 
 # Setting a default value for QT_API
-os.environ.setdefault(QT_API, 'pyqt5')
+os.environ.setdefault(QT_API, 'pyside2')
 API = os.environ[QT_API].lower()
 assert API in (PYQT5_API + PYQT4_API + PYSIDE_API + PYSIDE2_API)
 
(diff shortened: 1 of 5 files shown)


ccordoba12:
@diorcety, thanks for your contribution!

Could you better explain the rationale behind this change please?

diorcety:
By default pyside-uic convert C header field for custom widget in python import.
Cf https://github.com/pyside/Tools/blob/master/pysideuic/uiparser.py#L807

ccordoba12:
@diorcety, thanks for the explanation.

@astrofrog, do you approve this one?

astrofrog:
Is there an easy way to add a test for this?

ccordoba12:
@diorcety, please do find a way to add a test for this.

diorcety:
ok I wil try

ccordoba12:
@diorcety, thanks a lot for the test! Things look really good to me now.

@astrofrog, do you approve?

ccordoba12:
Pinging @astrofrog again. @astrofrog, I'd like to release 1.3 this week, so if you don't answer, we'll leave this for 1.4.

astrofrog:
Is the ``qcombobox_subclass.h`` file missing from here? If so, why does this not error?

diorcety:
The goal of this support is take a UI from a C++ project and use it in python project by changing h files by python file. qcombobox_subclass.h doesn't exist but qcombobox_subclass.py do.

PR #112, opened 2017-05-08
PR: Honor QStandardPaths location in Qt5

### LINKED ISSUE #89: QDesktopServices split into QDesktopServices and QStandardPaths
The QDesktopServices from Qt4 has been split into two namespaces, QDesktopServices and QStandardPaths, but there is no support for QStandardPaths in QtPy.  I would expect QtPy to support code like:

`home_location = QtCore.QStandardPaths.standardLocations(QtCore.QStandardPaths.HomeLocation)`

Qt4 API
http://doc.qt.io/qt-4.8/qdesktopservices.html

Qt5 API
http://doc.qt.io/qt-5/qdesktopservices.html
http://doc.qt.io/qt-5/qstandardpaths.html

rlaverde:
Fixes #89

The QDesktopServices from Qt4 has been split into two namespaces,
QDesktopServices and QStandardPaths, this allows a pyqt5 import of
QStandardPaths to work with pyqt4.

`from qtpy.QtCore import QStandardPaths`

I'm not sure if this is the best way to do it

### DIFF
--- a/qtpy/QtCore.py
+++ b/qtpy/QtCore.py
@@ -38,6 +38,26 @@ elif PYQT4:
                              QStringListModel)
     from PyQt4.QtCore import QT_VERSION_STR as __version__
 
+    # QDesktopServices has has been split into (QDesktopServices and
+    # QStandardPaths) in Qt5
+    # This creates a dummy class that emulates QStandardPaths
+    from PyQt4.QtGui import QDesktopServices as _QDesktopServices
+
+    class QStandardPaths():
+        StandardLocation = _QDesktopServices.StandardLocation
+        displayName = _QDesktopServices.displayName
+        DesktopLocation = _QDesktopServices.DesktopLocation
+        DocumentsLocation = _QDesktopServices.DocumentsLocation
+        FontsLocation = _QDesktopServices.FontsLocation
+        ApplicationsLocation = _QDesktopServices.ApplicationsLocation
+        MusicLocation = _QDesktopServices.MusicLocation
+        MoviesLocation = _QDesktopServices.MoviesLocation
+        PicturesLocation = _QDesktopServices.PicturesLocation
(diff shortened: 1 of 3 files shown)


goanpeca:
@rlaverde this looks like the easiest solution moving forward, I would say :+1:, @ccordoba12 ?

ccordoba12:
This is not exactly what's needed. Not all QDesktopServices of Qt4 were moved to QStandardPaths in Qt5.

You need to add to `qtpy.QtCore` only the objects that belong to QStandardPaths in Qt5, and remove them from `qtpy.QtGui`.

goanpeca:
@rlaverde now with can add tests that raise Import Error to check that everything works now as expected. Could you add those?

rlaverde:
>You need to add to qtpy.QtCore only the objects that belong to QStandardPaths in Qt5, and remove them from qtpy.QtGui.

It's not possible to delete objects that doesn't belong to the module

if a do, something like this
```
from PyQt4.QtGui import QDesktopServices as QStandardPaths
del QStandardPaths.openUrl
```
It get deleted from `QDesktopServices` because `QStandardPaths` is a reference to it; and won't be possible to import it from `QDesktopServices`

ccordoba12:
> from PyQt4.QtGui import QDesktopServices as QStandardPaths

This is what's wrong. I think we need to create a separate module called perhaps `_qstandardpaths`, import there all objects that must go there from PyQt4/PySide in Qt5, and then import this module as

    import _qstandardpaths as QStandardPaths

I think something like this should work.

rlaverde:
I tried creating a module, but `QDesktopServices` and `QStandardPaths` are classes, so I created dummy classes to allow its importing

rlaverde:
@ccordoba12 @goanpeca This is ready, could you review it 😄

goanpeca:
LGTM!

PR #117, opened 2017-05-15
PR: Add compatibility for the rename of qInstallMsgHandler to qInstallMessageHandler

### LINKED ISSUE #57: qInstallMessageHandler <-> qInstallMsgHandler
The installer for the message handler is called qInstallMsgHandler in PyQt4 and pyside and qInstallMessageHandler in PyQt5. In PyQt5 the message handler got an extra argument called context.

The reason to use qtpy is to have a program running in all three environments. Thus, there should be a unified way to use it (e.g. set context to None if pyqt4/pyside is used).

rlaverde:
Fixes #57


ccordoba12:
Since this is ready, we could include it in 1.3. @goanpeca, what do you say?

goanpeca:
Yes definitely!

PR #120, opened 2017-06-19
Revert "raise exception if old methods are called"

### LINKED ISSUE #119: Importing qtpy should not raise exceptions
Hi, qtpy is breaking code that does not even use it. This was already warned by @titusjan in https://github.com/spyder-ide/qtpy/pull/65#issuecomment-252472704

The following is a snippet where the problem is demonstrated. A PyQt4 program that does not use qtpy but imports it (perhaps it imports some other module that imports qtpy) gets broken. If the qtpy import is removed, the program reaches the `print 3` line, but if the import is done, an exception prevents reaching that line.

```python
import sip
sip.setapi('QString', 2)
sip.setapi('QVariant', 2)
sip.setapi('QDate', 2)
sip.setapi('QDateTime', 2)
sip.setapi('QTextStream', 2)
sip.setapi('QTime', 2)
sip.setapi('QUrl', 2)

from PyQt4 import QtGui, QtCore
app = QtGui.QApplication([])

w = QtGui.QHeaderView(QtCore.Qt.Vertical)
w.setResizeMode(w.Fixed)
print 1, w.setResizeMode

# Importing qtpy.Qwidgets changes w!!! (even if never using qtpy)
import qtpy.QtWidgets

print 2, w.setResizeMode
w.setResizeMode(w.Fixed)
print 3
```

IMHO, simply importing qtpy should never affect programs that do not use it (in the case of https://github.com/taurus-org/taurus/issues/401 , just importing `spyder.widgets.editor` breaks our module ). 

Would it be possible to avoid patching those methods? (or at least doing something less disruptive than raising an exception?)

Cheers!

cpascual:
Fix #119 by reverting commit 722e745932281a64ef64ac8dd189329bec462f00

### DIFF
--- a/qtpy/_patch/qheaderview.py
+++ b/qtpy/_patch/qheaderview.py
@@ -7,82 +7,48 @@
 
 def introduce_renamed_methods_qheaderview(QHeaderView):
 
-    _isClickable = QHeaderView.isClickable
     def sectionsClickable(self):
         """
         QHeaderView.sectionsClickable() -> bool
         """
-        return _isClickable(self)
+        return QHeaderView.isClickable(self)
     QHeaderView.sectionsClickable = sectionsClickable
-    def isClickable(self):
-        raise Exception('isClickable is only available in Qt4. Use '
-                        'sectionsClickable instead.')
-    QHeaderView.isClickable = isClickable
 
-
-    _isMovable = QHeaderView.isMovable
     def sectionsMovable(self):
         """
         QHeaderView.sectionsMovable() -> bool
         """
-        return _isMovable(self)
+        return QHeaderView.isMovable(self)
     QHeaderView.sectionsMovable = sectionsMovable
-    def isMovable(self):
-        raise Exception('isMovable is only available in Qt4. Use '
-                        'sectionsMovable instead.')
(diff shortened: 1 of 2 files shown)


ccordoba12:
@cpascual, I agree with your rationale for this PR, but instead of simply removing those exceptions, could you add a warning about those deprecated methods using the `warnings` module? Thanks!

cpascual:
No prob. I'll push as soon as I get it done.

ccordoba12:
Thanks! I'm sorry for not giving you feedback before about your work, 
but I hope we can merge it soon.

cpascual:
I pushed the solution with your suggested approach in a separate PR ( #121 ) in order not to "revert the reverted"

goanpeca:
@cpascual so should we close this one?

cpascual:
> @cpascual so should we close this one?

Yes, unless you want to reconsider it (see https://github.com/spyder-ide/qtpy/pull/121#issuecomment-318403455 ).

Actually, I wrote `closes #120` in the description of #121, so if you merge #121, this one should be [automagically closed](https://help.github.com/articles/closing-issues-using-keywords/)

PR #121, opened 2017-07-27
PR: Warn if QHeaderView deprecated methods are used

### LINKED ISSUE #119: Importing qtpy should not raise exceptions

cpascual:
This is a second attempt at fixing #119 according to the request from @ccordoba12 in https://github.com/spyder-ide/qtpy/pull/120#issuecomment-318371899 :
Warn instead of raising an exception if QHeaderView deprecated methods
are used. Also adapt unit tests accordingly.

Fixes #119
Closes #120

### DIFF
--- a/qtpy/_patch/qheaderview.py
+++ b/qtpy/_patch/qheaderview.py
@@ -4,6 +4,7 @@
 #
 # Licensed under the terms of the MIT License
 # (see LICENSE.txt for details)
+import warnings
 
 def introduce_renamed_methods_qheaderview(QHeaderView):
 
@@ -15,8 +16,9 @@ def introduce_renamed_methods_qheaderview(QHeaderView):
         return _isClickable(self)
     QHeaderView.sectionsClickable = sectionsClickable
     def isClickable(self):
-        raise Exception('isClickable is only available in Qt4. Use '
-                        'sectionsClickable instead.')
+        warnings.warn('isClickable is only available in Qt4. Use '
+                        'sectionsClickable instead.', stacklevel=2)
+        return _isClickable(self)
     QHeaderView.isClickable = isClickable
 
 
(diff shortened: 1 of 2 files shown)


cpascual:
Just one disclaimer: I am personally more in favour of the solution in #120 than this one because with this one importing qtpy still does have a side effect for modules not using it (which IMHO is harmful).

That said, I think that the side effect in this case is very unlikely to cause trouble (certainly much less than raising exceptions!) and therefore I see it as a reasonable compromise if you do not like #120.

ccordoba12:
Thanks for your patience @cpascual! This is going in :-)

PR #122, opened 2017-07-31
PR: Patch QFiledialog and offer new methods

### LINKED ISSUE #26: Get rid of getexistingdirectory and patch QFileDialog?
I liked the approach of py.qode better cause it did not relly on helper functions, but instead it patched QFileDialog to get the correct behavior.

We could do the same and get rid of the helper functions.

@ccordoba12 ?

goanpeca:
Fixes #26

### DIFF
--- a/qtpy/_patch/qfiledialog.py
+++ b/qtpy/_patch/qfiledialog.py
@@ -1,0 +1,159 @@
+# -*- coding: utf-8 -*-
+"""QFileDialog patch."""
+
+import sys
+
+from .py3compat import to_text_string, is_text_string
+
+
+def patch_filedialog(QFileDialog):
+    """Patch QFile Dialog."""
+
+    def _qfiledialog_wrapper(attr, parent=None, caption='', basedir='',
+                             filters='', selectedfilter='', options=None):
+        if options is None:
+            options = QFileDialog.Options(0)
+        try:
+            # PyQt <v4.6 (API #1)
+            from .QtCore import QString
+        except ImportError:
+            # PySide or PyQt >=v4.6
+            QString = None  # analysis:ignore
+        tuple_returned = True
+        try:
+            # PyQt >=v4.6
+            func = getattr(QFileDialog, attr+'AndFilter')
+        except AttributeError:
+            # PySide or PyQt <v4.6
+            func = getattr(QFileDialog, attr)
+            if QString is not None:
+                selectedfilter = QString()
+                tuple_returned = False
+
+        # Calling QFileDialog static method
+        if sys.platform == "win32":
+            # On Windows platforms: redirect standard outputs
+            _temp1, _temp2 = sys.stdout, sys.stderr
+            sys.stdout, sys.stderr = None, None
+        try:
+            result = func(parent, caption, basedir,
+                          filters, selectedfilter, options)
+        except TypeError:
+            # The selectedfilter option (`initialFilter` in Qt) has only been
+            # introduced in Jan. 2010 for PyQt v4.7, that's why we handle here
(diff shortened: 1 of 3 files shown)


goanpeca:
@Nodd better late than never. So this patch exposes the methods with Qt case, plus a 2, because they change the return values. So changing compatibility for existing code, is just adding a 2 and accepting a tupple instead of a single argument.

Thoughts @ccordoba12 ?

qtpy/_patch/qfiledialog.py
```diff
@@ -0,0 +1,159 @@
...
+                sys.stdout, sys.stderr = _temp1, _temp2
+        if not is_text_string(result):
+            # PyQt API #1
+            result = to_text_string(result)
+        return result
+
+    QFileDialog.getOpenFileName2 = _getOpenFileName
+    QFileDialog.getOpenFileNames2 = _getOpenFileNames
+    QFileDialog.getSaveFileName2 = _getSaveFileName
+    QFileDialog.getExistingDirectory2 = _getExistingDirectory
```
ccordoba12: Do these names exists in Qt? I mean, `getOpenFileName2`, etc?

ccordoba12:
Tests are failing.

goanpeca:
>Tests are failing.

Wow really??!?! thanks for letting me know 😈

ccordoba12:
Jeje, just in case you haven't noticed it ;-p

qtpy/_patch/qfiledialog.py
goanpeca: No they don't, but if we replace the normal methods, the return is changed, so it would be weird.

We can warn QFileDialog.getExistingDirectory to better use QFileDialog.getExistingDirectory2. 

Unfortunately I cant run any test on this, cause I am either stuch in the exec loop or segfaulting. I think that in general is a very bad idea/bad practice to use this methods that call exec/modal directly. It is better and safer (as seen by the latest fixes of @rlaverde in spyder) to avoid using these methods :-\

goanpeca: `RuntimeError: no access to protected functions or signals for objects not created from Python` there is no way to add tests for this PR cause testing methods that call exec block testing, even if I try to access them by name, I will not be able to call the methods.

goanpeca:
Yeah, this is definitely not testable. I would say we should not really encourage the use of the compat funcs. And we should warn the user when using methods that do not create python objects the user can access or interact with (QMessageBox and QFileDialog methods)

ccordoba12:
> Yeah, this is definitely not testable

Ok, no problem with that.

However, I don't agree too much with the addition of warnings in this PR because we will have to add code in Spyder to swallow them.

goanpeca:
>However, I don't agree too much with the addition of warnings in this PR because we will have to add code in Spyder to swallow them.

Yes, I can remove it for now. My point is that we should move away from using those methods in Spyder as we cannot test them plus they sometimes lead to odd segfaults. It is always better to create the object first and then exec it (but we have the reference to the python object and can interact with it if needed on tests or avoid segfaults) like https://github.com/spyder-ide/spyder/pull/4870/files#diff-0308a6d064ae2a204bd97de4f832e1f4R1430

ccordoba12:
@goanpeca, could you finish this PR?

cpascual:
I personally think that qtpy should not patch the imported Qt objects. The rationale is the same as I already expressed in #119 : side effects for people not even using qtpy should be avoided at all costs.

For example, with this PR an application using perfectly valid PyQt4 code could break just because it imported a widget from spyder

ccordoba12:
Ok, let's close this for now due to @cpascual's remarks and this being stalled.

PR #128, opened 2017-08-17
PR: Add QtHelp Wrapper

### LINKED ISSUE #127: Add Wrapper for QtHelp [feature request]
no problem importing QtHelp from PyQt5, but cannot import from qtpy.

No problem:

`from PyQt5 import QtHelp`

Module not found:

`from qtpy import QtHelp`

rlaverde:
Fixes #127

### DIFF
--- a/qtpy/QtHelp.py
+++ b/qtpy/QtHelp.py
@@ -1,0 +1,24 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © 2009- The Spyder Development Team
+#
+# Licensed under the terms of the MIT License
+# (see LICENSE.txt for details)
+
+"""QtHelp Wrapper."""
+
+from . import PYQT5
+from . import PYQT4
+from . import PYSIDE
+from . import PYSIDE2
+
+if PYQT5:
+    from PyQt5.QtHelp import *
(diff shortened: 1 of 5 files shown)


qtpy/QtHelp.py
```diff
@@ -0,0 +1,24 @@
...
+from . import PYQT4
+from . import PYSIDE
+from . import PYSIDE2
+
+if PYQT5:
+    from PyQt5.QtHelp import *
+elif PYSIDE2:
+    # Current wheels don't have this module
+    # from PySide2.QtHelp
+    pass
```
goanpeca: But, this would result in a successful import. Should we at least add a warning?

@ccordoba12 ?

rlaverde: If we are going to add a warning we should add it to qtmultimedia too

ccordoba12: Yeah, we should add warning to this one and also QtMultimedia.

rlaverde:
I created a custom warning? is that ok? or should I just use the standard warning ` warnings.warn()` as #121

ccordoba12:
> I created a custom warning? is that ok?

I think it's ok, thanks!

PR #136, opened 2017-11-15
PR: Add QtSql wrapper (incl. test)

### LINKED ISSUE #135: Add Wrapper for QtSql [feature request]
I guess/hope the implementation for QtSql is pretty straight forward and similar to QtSvg. Thanks in advance.

daniel-edler:
Fixes #135 

----

Fix (my feature request) #135

### DIFF
--- a/qtpy/QtSql.py
+++ b/qtpy/QtSql.py
@@ -8,11 +8,7 @@
 """Provides QtSql classes and functions."""
 
 # Local imports
-from . import PYQT4
-from . import PYSIDE2
-from . import PYQT5
-from . import PYSIDE
-from . import PythonQtError
+from . import PYQT5, PYSIDE2, PYQT4, PYSIDE, PythonQtError
 
 if PYQT5:
     from PyQt5.QtSql import *
--- a/qtpy/QtSql.py
+++ b/qtpy/QtSql.py
@@ -5,7 +5,7 @@
 # Licensed under the terms of the MIT License
 # (see LICENSE.txt for details)
 # -----------------------------------------------------------------------------
-"""Provides QtSvg classes and functions."""
+"""Provides QtSql classes and functions."""
 
 # Local imports
 from . import PYQT4
--- a/qtpy/QtSql.py
+++ b/qtpy/QtSql.py
@@ -1,0 +1,28 @@
+# -*- coding: utf-8 -*-
+# -----------------------------------------------------------------------------
+# Copyright © 2009- The Spyder Development Team
+#
+# Licensed under the terms of the MIT License
+# (see LICENSE.txt for details)
+# -----------------------------------------------------------------------------
+"""Provides QtSvg classes and functions."""
+
+# Local imports
+from . import PYQT4
+from . import PYSIDE2
+from . import PYQT5
+from . import PYSIDE
+from . import PythonQtError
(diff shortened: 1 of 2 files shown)


goanpeca:
@daniel-edler thanks :-), did you check if QtSql change between qt 4.8 and 5.6 (what we currently support?)

daniel-edler:
If I get your question right, the [qt5 documentation](http://doc.qt.io/qt-5/qtsql-module.html) lists the same components as the [qt4.8 documentation](http://doc.qt.io/qt-4.8/qtsql-module.html)

qtpy/QtSql.py
```diff
@@ -0,0 +1,28 @@
...
+# -----------------------------------------------------------------------------
+# Copyright © 2009- The Spyder Development Team
+#
+# Licensed under the terms of the MIT License
+# (see LICENSE.txt for details)
+# -----------------------------------------------------------------------------
+"""Provides QtSql classes and functions."""
+
+# Local imports
+from . import PYQT4
```
rlaverde: This could be just one import:

`from . import PYQT5, PYSIDE2, PYQT4, PYSIDE, PythonQtError`

qtpy/tests/test_qtsql.py
```diff
@@ -0,0 +1,23 @@
...
+import pytest
+from qtpy import QtSql
+
+def test_qtsql():
+    """Test the qtpy.QtSql namespace"""
+    assert QtSql.QSqlDatabase is not None
+    # assert QtSql.QSqlDriverCreator is not None
+    assert QtSql.QSqlDriverCreatorBase is not None
+    assert QtSql.QSqlDriver is not None
+    #assert QtSql.QSqlDriverPlugin is not None
```
rlaverde: You could move this comment out lines to another test skipped in PYSIDE2

```
from qtpy import PYSIDE2

@pytest.mark.skipif(PYSIDE2, reason="It fails on PySide2")
def test_qtsql_no_pyside():
    assert QtSql.QSqlDriverCreator is not None
    assert QtSql.QSqlDriverPlugin is not None
```

qtpy/QtSql.py
daniel-edler: I personally prefer the way as it is in my commit. But it's not my project and if it's common here there's no point of argue => it will be in my next update

qtpy/tests/test_qtsql.py
daniel-edler: I don't think so. It is not part of my PyQt5 environment and the checks told me that it is also not the case for the other three case. Actually for the QSqlDriverCreater class/1st commit the test succeeded for PySide2 because there I excluded PySide2 completely. (Too much copy'n'paste...) But I doubt that it is included there

rlaverde: I understood, It's ok to leave if comment out, maybe add a comment explaining that those features aren't yet  present in the Qt API

qtpy/QtSql.py
rlaverde: I'm not sure, I think there's code styles guidelines for qtpy, @goanpeca ?

ccordoba12:
Thanks @daniel-edler for your contribution and @rlaverde for your review.

PR #139, opened 2018-02-13
PR: If a Qt binding is already imported, then use it.

### LINKED ISSUE #138: If one binding has already been imported, then qtpy should just use it
Consider, for example, a system where solely pyqt4 and pyqt5 are installed.  If one does
```
import PyQt5.QtGui
import qtpy
```
this fails with
```
qtpy.PythonQtError: No Qt bindings could be found
```
(even though obviously there is a Qt binding available) because PyQt4.QtGui cannot be imported at all after PyQt5.QtGui has already been imported.

Fortunately there is an easy solution, which is to check, before doing anything else, whether PyQt4/5/PySide/2 are already present in sys.modules, and, if one of them is, just use it.

Real use case:
matplotlib (which can use any of the bindings) is considering switching from using its own qt shim to qtpy.  However, it is not rare that one user specifies a given binding to use (via the matplotlibrc backend param), but sometimes wants to use some GUI program that is explicitly written (solely) against PyQt5.  Then starting the GUI will import PyQt5 and matplotlib would fail to initialize its backend if it did not first check whether a given binding had already been forced for the process.

anntzer:
Fixes #138.

See #138 for rationale.

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -10,12 +10,13 @@
 **QtPy** is a shim over the various Python Qt bindings. It is used to write
 Qt binding indenpendent libraries or applications.
 
-The shim will automatically select the first available API (PyQt5, PyQt4 and
-finally PySide).
+If one of the APIs has already been imported, then it will be used.
 
-You can force the use of one specific bindings (e.g. if your application is
-using one specific bindings and you need to use library that use QtPy) by
-setting up the ``QT_API`` environment variable.
+Otherwise, the shim will automatically select the first available API (PyQt5,
+PySide2, PyQt4 and finally PySide); in that case, you can force the use of one
+specific bindings (e.g. if your application is using one specific bindings and
+you need to use library that use QtPy) by setting up the ``QT_API`` environment
+variable.
 
 PyQt5
 =====
@@ -62,6 +63,7 @@ packages::
 """
 
 import os
+import sys
 
 # Version of QtPy
 from ._version import __version__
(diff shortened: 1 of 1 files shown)


goanpeca:
Looks good, some test failing though. Any chance we could write a test for this?

anntzer:
Properly testing this looks slightly tricky as it appears that currently you have four CI setups each with just one of the dependencies installed, but to test this feature meaningfully you'd need another setup with two of them installed.

We've been seeing failing tests (segfaulting) with the recently released pyqt5.10 on matplotlib as well and have downgraded the CI to use pyqt5.9.  Do you know whether qtpy master passes with pyqt5.10?

ccordoba12:
> Properly testing this looks slightly tricky

I agree. Since this change is quite minimal and easy to understand, I'm ok to not add a test for this one. Besides, Matplotlib is going to be the real test for this addition.

 @goanpeca?

> We've been seeing failing tests (segfaulting) with the recently released pyqt5.10 on matplotlib as well and have downgraded the CI to use pyqt5.9

Ok, good to know. Let me pin pyqt to 5.9 here too, so our tests can pass @anntzer.

ccordoba12:
@anntzer, please rebase against master to get the fix for our tests.

anntzer:
done

ccordoba12:
Thanks @anntzer for this. Being a dependency of Matplotlib is a huge achievement for a small project like this one!

I'll try to release 1.4 as soon as possible (a couple of weeks at most).

anntzer:
It's not a dependency yet but may become one soon... (https://github.com/matplotlib/matplotlib/issues/6118)

ccordoba12:
Yep, I know.

ccordoba12:
@anntzer, qtpy 1.4 is out!

anntzer:
thanks for the note :)

PR #156, opened 2018-05-26
New behavior when start import QtPy - setting Qt binding

dpizetta:
This PR restructure the mainly the init file, implementing new functions to help QtPy to get info about Qt API's. 

- QtPy will always try 4 times (4 API's) if None is found.
- Add and improve warnings if changes and fails occurs.
- Help functions may be part of scripts to help developers in the future, ex.:

```
qtpy --available_api >> PySide, PyQt5
qtpy --api_info=PyQt5 >> APIVersion, QTVersion ...
```

More details about the new behavior are described in init documentation.

It passed tests locally, but some variables were changed in init file, that someone is using somewhere in other projects.

`PYSIDE_VERSION > API_VERSION`

Feel free to criticise and comment the code.


dpizetta:
About CI errors, I can change the use of importlib.utils to try...import to avoid problems with py27. Waiting for your reviews :) Tks

goanpeca:
`PYSIDE_VERSION > API_VERSION`

Those changes might be problematic, we would need to preserve names, and maybe add deprecation notices if we want to change this for 1.5, or even 2.x ?

@ccordoba12 ?

qtpy/__init__.py
```diff
@@ -26,42 +63,42 @@
...
-======
+=======
 
-Set the QT_API environment variable to 'pyside2' before importing other
+Set the QT_API environment variable to 'PySide2' before importing other
 packages::
 
     >>> import os
-    >>> os.environ['QT_API'] = 'pyside2'
+    >>> os.environ['QT_API'] = 'PySide2'
```
ccordoba12: Sorry but I don't have much time right now to do a thorough review of this. For now, please revert this change you made here about using capitalization to set `QT_API`.

There's a well established convention among Python projects that deal with different Qt bindings to use `QT_API` as we are using it here, so please revert that.

ccordoba12:
@dpizetta, I left an initial comment and also please fix the failures in our CIs (the best way you consider it) before we can start to review your work.

qtpy/__init__.py
```diff
@@ -251,38 +274,39 @@ def get_api_information(api_name):
...
     else:
-        if api_version and qt_version:
-            API = api_name
-            API_VERSION = api_version
-            QT_VERSION = qt_version
-            if API == 'PyQt4':
+        if API_VERSION and QT_VERSION:
+            API = api_name.lower()
+            API_NAME = api_name
+            if api_name == 'PyQt4':
```
goanpeca: Need to fix these values not in lowercase?

dpizetta: hi @goanpeca , inside the code, I'm using the importing name because we can pass it to the functions to import. The lowercase is in API.

In the future, a better idea to implement the function get_api_information(api_name) is to use pgkutil and/or importlib. When you use try...import, all the modules that you have tried are in sys.modules. And both libs accept a string to make the temporary import. I didn't do this way now because I need to make sure of what implications this can make [here](https://github.com/spyder-ide/qtpy/pull/156/files#diff-bca24c6a427bc0a1c74aa972ec74d9f6R172)

dpizetta: The CI problem requires changes in test also because the order for attempts (PyQt5, PySide2, etc..) have changed. I'm working on it.

dpizetta:
The fail in CI for PySide2 maybe is caused by the low speed to download the wheels. I have this problem here too. The latest version is sadly very low to download. I have used latest-1 :) Maybe enable cache can help if not enabled.

dpizetta:
The problems with pyside2 servers seem to be solved... https://bugreports.qt.io/browse/PYSIDE-684 :) I will try to recheck.

ccordoba12:
@dpizetta, please rebase this one in top of master to fix the error in CircleCI.

goanpeca:
I forgot about this. Will review it :-)

dpizetta:
Hi guys, I would like to ask both of you @goanpeca and @ccordoba12  to wait for me to update this PR. I've also more ideas that can help this PR. Tks

goanpeca:
Ok @dpizetta thanks for working on this and sorry for the late review! Please let us know

goanpeca:
One genera comment is that if you want to use the QT_API to import something, I think is better to let QT_API be case independent (pyside2, PySide2 etc.. should do the same thing), and have a map somewhere to use for your needs

```python
QT_API = 'PySide2'
MAP = {
    'pyside2': 'PySide2',
    ...
}
qt_api_module = MAP.get(QT_API.lower())
```

ccordoba12:
That's the current model, i.e. we accept bindings in upper or lower case names (see PR #45). @dpizetta, if that's changed by your PR, please revert it.

ccordoba12:
@dpizetta, any updates on this one?

dpizetta:
@ccordoba12 , sorry I forgot about this one. Can you give me one more week? I put in my schedule. Sorry for the delay.

ccordoba12:
Ok, thanks! There's no rush in our side (I mean, we don't need this for Spyder), but I think it's really necessary for the next release of QDarkstyle, right?

dpizetta:
When finished I'll ping you for a review. Tks

dpizetta:
Can you answer me a question? About this code: https://github.com/spyder-ide/qtpy/blob/0aec37a98c740a026e3088aa06d7bb38b467ae8f/qtpy/__init__.py#L117

Actually if FORCE_QT_API is set, you want that the used Qt binding is given by the previous imported binding? 

If I do, imaging that I have both PySide and PyQt5 installed:

```
import os
import PySide2
os.environ['FORCE_QT_API']=True
os.environ['QT_API']='pyqt5'
import qtpy
```
You want that PySide2 is used and if:

```
import os
import PySide2
os.environ['QT_API']='pyqt5'
import qtpy
```
You want that PyQt5 is used. Is it that? For me, it is kind confusing the name FORCE_QT_API and the result. Tks

ccordoba12:
> Actually if FORCE_QT_API is set, you want that the used Qt binding is given by the previous imported binding?

The idea is that if FORCE_QT_API is set, then qtpy must respect the binding previously set in QT_API instead of switching to the currently imported binding. That allows applications that want to use PySide2 and PyQt5 at the same time, as in your second example above.

dpizetta:
> The idea is that if FORCE_QT_API is set, then qtpy must respect the binding previously set in QT_API instead of switching to the currently imported binding. That allows applications that want to use PySide2 and PyQt5 at the same time, as in your second example above.

Oh, ok. I think it is not a good idea to do this (in case of merging codes like putting a PyQt5 widget into a PySide2 window) but I'm ok. I'll propose some messages of warning to try to get all problematic possible cases, then we can discuss. For me, the code is doing the opposite, but I'll make some tests here (for me it should be "is None"). 

```
# When `FORCE_QT_API` is set, we disregard
# any previously imported python bindings.
if os.environ.get('FORCE_QT_API') is not None:
    if 'PyQt5' in sys.modules:
        API = initial_api if initial_api in PYQT5_API else 'pyqt5'

```

Tks, @ccordoba12 .

dpizetta:
Also, any ideas to solve the CI problems? I don't get it why the code is not passing in that part if I'm using same as the master branch.

```

>       assert '10.11' in str(e.value)
E       assert '10.11' in "LooseVersion instance has no attribute 'version'"
E        +  where "LooseVersion instance has no attribute 'version'" = str(AttributeError("LooseVersion instance has no attribute 'version'",))
E        +    where AttributeError("LooseVersion instance has no attribute 'version'",) = <ExceptionInfo AttributeError tblen=4>.value

qtpy/tests/test_macos_checks.py:82: AssertionError
```

ccordoba12:
> For me, the code is doing the opposite, but I'll make some tests here (for me it should be "is None")

Ok, thanks for thinking about it.

> Also, any ideas to solve the CI problems?

It seems there's a problem getting `e.value` in that test.

dpizetta:
Hi guys, any ideas for [this problem](https://ci.appveyor.com/project/spyder-ide/qtpy/builds/26532575/job/471xto8c0kglaarc#L261) in CI? Tks

Also, is there any problem setting debug level for the logging in test mode? I think we get more info from the code - I put some in __init__.

dpizetta:
@goanpeca could give some help with those issues on CI's, both are in Python2.7 and related to the return of a function that returns -1 when compared to 1 (-1 == 1). Thanks. Later I'll solve the conflicts, so you can review the code. It should be working without any drastic API changes.

goanpeca:
>@goanpeca could give some help with those issues on CI's, both are in Python2.7 and related to the return of a function that returns -1 when compared to 1 (-1 == 1). 

Sure! will do :-)

dpizetta:
@goanpeca some new about this one? I think it is working properly, just the problem with the tests that seem not to be with the implementation here. If I could help more, please, let me know tks

PR #159, opened 2018-06-03
Install Pyside2 from conda-forge

jschueller:


ccordoba12:
I don't agree with this because it implies some regressions in our tests. Perhaps we should move the official Qt packages to another location that doesn't have download issues.

ccordoba12:
> What regressions are you talking about ?

Sorry, there are not regressions, but I saw you're commenting tests that are passing right now.

ccordoba12:
So I think it's better to download the official Qt PySide2 wheels and put them in Bintray or something, the change the URL here, and that'll solve the failures in CIrcleCI.

dpizetta:
I think this problem with pyside2 servers are solved... https://bugreports.qt.io/browse/PYSIDE-684 :)

jschueller:
cool

PR #160, opened 2018-06-11
Fix pyside2 wheels install

jschueller:


jschueller:
could someone restart the ci ?

PR #171, opened 2018-08-30
Allow py3compat.py to pass flake8 tests

cclauss:
Other projects are including qtpy in their repos but if they run [__flake8__](http://flake8.pycqa.org) on their repos then qtpy/py3compat.py fails.  These proposed changes solve that problem using two techniques:
1. Python porting best practice [__use feature detection instead of version detection__](https://docs.python.org/3/howto/pyporting.html#use-feature-detection-instead-of-version-detection) to avoid _undefined names_
2. _# noqa F401_ to suppress _x imported but not used_ messages

### DIFF
--- a/qtpy/py3compat.py
+++ b/qtpy/py3compat.py
@@ -18,8 +18,8 @@ This module should be fully compatible with:
 
 from __future__ import print_function
 
-import sys
 import os
+import sys
 
 PY2 = sys.version[0] == '2'
 PY3 = sys.version[0] == '3'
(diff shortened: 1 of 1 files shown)


cclauss:
The circleci build failures seem to be unrelated to these changes.

ccordoba12:
> Other projects are including qtpy

Which projects? And why not these projects simple declare `qtpy` in their setup.py?

cclauss:
https://github.com/git-cola/git-cola

ccordoba12:
Well, if you want to embed qtpy in your project, I'm fine with it. However, please don't complain upstream for they don't following your code quality practices.

Can't you just skip that file from flake8?

cclauss:
It is not my project.  I am with you.  Putting qtpy in setup.py is the way to go.  And yes: https://github.com/git-cola/git-cola/pull/879/files#diff-b67911656ef5d18c4ae36cb6741b7965R274

PR #176, opened 2018-12-03
PR: Don't warn about bindings change if user did not specify a binding

ignamv:


ccordoba12:
What's the purpose of this PR? Please be more specific.

ignamv:
If the user doesn't specify a binding, qtpy tries to use pyqt5. If pyqt5 is not found and pyside is used, qtpy produces a warning. I believe this warning should only happen if the user explicitly requested a specific binding. If no binding was specified, it's safe to say he doesn't care which binding is used. Also, it is arbitrary to throw a warning with pyside and not with pyqt5.

ccordoba12:
Note: The Circle failures are unrelated to this.

ignamv:
No problem, thank you for qtpy!

PR #182, opened 2019-01-25
PR: Prevent warnings for equivalent APIs

FedeMiorelli:
Situation: user has already imported PyQt4 and QT_API is set to pyqt.
qtpy will issue a spurious warning complaining that initial_api pyqt is different from selected one pyqt4, even if they are the same.

This code will make sure that in case of pre-loaded QT modules, the API is reset to the same initial_api if compatible, otherwise to the default.

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -113,13 +113,13 @@ PYQT4 = PYSIDE = PYSIDE2 = False
 
 
 if 'PyQt5' in sys.modules:
-    API = 'pyqt5'
+    API = initial_api if initial_api in PYQT5_API else 'pyqt5'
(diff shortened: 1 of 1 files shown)


ccordoba12:
@FedeMiorelli, thanks for your contribution!

@dpizetta, what do you think about this one?

dpizetta:
Hi, @FedeMiorelli  tks for the catch! I'd mention that this case only occurs to the pyqt4, that accept to names, but as we are setting a list for the possible names, you code make the comparison generalized. I'm OK.

FedeMiorelli:
Thank you both, HTH.
Cheers,
Federico

PR #185, opened 2019-02-28
Improved "cannot find API" error message

Erotemic:
I encountered an error when trying to use the `labelme` software. It was totally my fault, but I would have been able to figure that out sooner if the error message was more clear. 

It turned out that I had the `export QT_API=pyqt` environment variable in my bashrc (which I must have put in years ago and forgot about it). However, I have a modern python3 + pyqt5 environment setup. The init file tried to resolve to PyQt4. The error message I got was: "No Qt bindings could be found", which was confusing because I had pyqt5 installed. It also spit out an error about pyside because it tried to fall back on that. 

Unfortunately, because the PyQt4 code is after the PyQt5 code, it never tried to fallback on pyqt5 (another potential place for improvement).  

I've updated the init to warn each time it fails to fallback on a different API, and I also updated the final error message to indicate if the user has explicitly set the QT_API environment variable (and suggest that they may want to change it). 

Anyway, I hope this helps any other unfortunate soul that finds themself in my situation.

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -100,12 +100,21 @@ PYSIDE_API = ['pyside']
 # Names of the expected PySide2 api
 PYSIDE2_API = ['pyside2']
 
+VALID_APIS = PYQT5_API + PYQT4_API + PYSIDE_API + PYSIDE2_API
+
+# Record if an explicit version of Qt was requested, so we can report this if
+# an error occurs.
+_EXPLICIT_API = None
+if QT_API in os.environ:
+    _EXPLICIT_API = os.environ[QT_API]
+
 # Setting a default value for QT_API
 os.environ.setdefault(QT_API, 'pyqt5')
 
 API = os.environ[QT_API].lower()
 initial_api = API
-assert API in (PYQT5_API + PYQT4_API + PYSIDE_API + PYSIDE2_API)
+
+assert API in VALID_APIS
 
 is_old_pyqt = is_pyqt46 = False
 PYQT5 = True
@@ -120,6 +129,9 @@ elif 'PyQt4' in sys.modules:
     API = 'pyqt4'
 elif 'PySide' in sys.modules:
     API = 'pyside'
+
+
+_fallback_fmt = 'The binding "{}" could not be found, try fallback "{}"'
 
 
 if API in PYQT5_API:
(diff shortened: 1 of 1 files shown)


goanpeca:
@Erotemic thanks for working on this!

ccordoba12:
@Erotemic, thanks for your contribution! This now needs a merge with master to fix a merge conflict.

qtpy/__init__.py
```diff
@@ -100,12 +100,21 @@ class PythonQtWarning(Warning):
 # Names of the expected PySide2 api
 PYSIDE2_API = ['pyside2']
 
+VALID_APIS = PYQT5_API + PYQT4_API + PYSIDE_API + PYSIDE2_API
+
+# Record if an explicit version of Qt was requested, so we can report this if
+# an error occurs.
+_EXPLICIT_API = None
+if QT_API in os.environ:
+    _EXPLICIT_API = os.environ[QT_API]
```
goanpeca: Are there any other variables names that could have been used? or is this exclusive?

Erotemic: I'm don't quite understand the question, could you rephrase?

goanpeca: @Erotemic what I wat to say if QT Docs only talk about `QT_API` as an environment variable, or are there other variants?

ccordoba12: I think `QT_API` was defined by the Python community (probably by Matplotlib or IPython).

goanpeca: Hmm ok :-), then it should be fine

goanpeca:
Ready to merge @ccordoba12 ?

qtpy/__init__.py
```diff
@@ -148,7 +162,9 @@ class PythonQtWarning(Warning):
 
             del macos_version
     except ImportError:
-        API = os.environ['QT_API'] = 'pyside2'
+        FALLBACK_API = 'pyside2'
+        warnings.warn(_fallback_fmt.format(API, FALLBACK_API), RuntimeWarning)
```
ccordoba12: I don't agree with displaying a warning for each binding because it'd make importing qtpy very verbose.

Instead, I think we should do this only if `_EXPLICIT_API is not None` and warn that a new binding `foo` is tried because `_EXPLICIT_API` couldn't be found.

qtpy/__init__.py
```diff
@@ -205,7 +225,16 @@ class PythonQtWarning(Warning):
...
     except ImportError:
-        raise PythonQtError('No Qt bindings could be found')
+        warnings.warn('The binding "{}" could not be found, '
+                      'cannot fall back'.format(API), RuntimeWarning)
+        if _EXPLICIT_API:
+            raise PythonQtError((
+                'No Qt bindings could be found for {}={}, which was '
+                'explicitly requested. Try unsetting the {!r} environ '
+                'or explicitly setting it to a different valid API: {!r}'
+            ).format(QT_API, _EXPLICIT_API, QT_API, VALID_APIS))
```
ccordoba12: With my suggested change above, I think there's no need to raise this error.

ccordoba12:
Closing due to lack of response.

PR #191, opened 2019-06-11
PR: Add Qt 3D bindings

### LINKED ISSUE #172: Support for Qt3D
I assume Qt3D is not wrapped as for PyQt5 the bindings are distributed as a separate package. Is this the case? Is there any plan to eventually wrap Qt3D or other separate components like QtDataVisualization?

jschueller:
Closes #172


jschueller:
Not everything is installable on CI's, so tests might be skipped.

qtpy/Qt3DAnimation.py
```diff
@@ -0,0 +1,22 @@
...
+from . import PYQT5, PYSIDE2, PythonQtError
+
+if PYQT5:
+    from PyQt5.Qt3DAnimation import *
+elif PYSIDE2:
+    # https://bugreports.qt.io/projects/PYSIDE/issues/PYSIDE-1026
+    import PySide2.Qt3DAnimation as __temp
+    import inspect
+    for __name in inspect.getmembers(__temp.Qt3DAnimation):
+        globals()[__name[0]] = __name[1]
```
ccordoba12: This looks very odd, although I understand it's a bug in PySide2. Would you be willing to change it when the bug is fixed upstream?

jschueller: Yes, of course, I opened the issue so I'll get notified if anything moves upstream.

ccordoba12: Ok, that's all I wanted to know. Thanks!

ccordoba12:
I just released version 1.8.0 with these changes.

jschueller:
Thanks!

PR #194, opened 2019-07-05
PR: Add FORCE_QT_API environment variable

### LINKED ISSUE #192: Binding Selection Logic
Hi,

In reference to [this section of code](https://github.com/spyder-ide/qtpy/blob/master/qtpy/__init__.py#L118), I was wondering what the reason is for the logic that makes the binding selection ignore the env var when a binding has already been imported.

In our situation, we're integrating tools into existing applications for the purpose of creating VFX.  These applications are built with Qt4/5, and some ship with python bindings - mostly PySide but some with PyQt.  Our tools are targeting PySide as that's an agreed upon industry standard and as such we've set the QT_API env var to be one of the pyside variants.

A problem arises however, when the host application ships with PyQt and imports it prior to our code importing qtpy.  The current behavior at the linked code ignores the env var in this case and falls back to using the lib that was already imported.

In our case, it's perfectly fine to have PySide2 co-existing with PyQt5 as our integrations aren't required to be parented to the PyQt5 widgets.

It's also worth noting that this need has come up through our use of qtawesome which implicitly uses qtpy as a shim layer.  The VFX industry has a defacto relationship with Qt.py to provide this behavior and we've subsequently made a patch to qtawesome to import Qt through that library for the time being.

This issue could be solved in either the qtawesome or the qtpy project, but I thought I'd start the discussion here as patching this logic seems like a smaller cut than trying to provide a shim for the shims in qtawesome.

Thanks,
Jase

JasonGilholme:
In response to this thread:

https://github.com/spyder-ide/qtpy/issues/192

-J

----

Fixes #192.

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -114,15 +114,17 @@ is_old_pyqt = is_pyqt46 = False
 PYQT5 = True
 PYQT4 = PYSIDE = PYSIDE2 = False
 
-
-if 'PyQt5' in sys.modules:
-    API = initial_api if initial_api in PYQT5_API else 'pyqt5'
-elif 'PySide2' in sys.modules:
-    API = initial_api if initial_api in PYSIDE2_API else 'pyside2'
-elif 'PyQt4' in sys.modules:
-    API = initial_api if initial_api in PYQT4_API else 'pyqt4'
-elif 'PySide' in sys.modules:
-    API = initial_api if initial_api in PYSIDE_API else 'pyside'
+# When `FORCE_QT_API` is set to "1", we disregard
+# any previously imported python bindings.
+if os.environ.get('FORCE_QT_API') == '1':
+    if 'PyQt5' in sys.modules:
+        API = initial_api if initial_api in PYQT5_API else 'pyqt5'
+    elif 'PySide2' in sys.modules:
+        API = initial_api if initial_api in PYSIDE2_API else 'pyside2'
+    elif 'PyQt4' in sys.modules:
+        API = initial_api if initial_api in PYQT4_API else 'pyqt4'
+    elif 'PySide' in sys.modules:
+        API = initial_api if initial_api in PYSIDE_API else 'pyside'
 
 
 if API in PYQT5_API:
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -114,9 +114,9 @@ is_old_pyqt = is_pyqt46 = False
 PYQT5 = True
 PYQT4 = PYSIDE = PYSIDE2 = False
 
-# When `FORCE_QT_API` is set to "1", we disregard
+# When `FORCE_QT_API` is set, we disregard
 # any previously imported python bindings.
-if os.environ.get('FORCE_QT_API') == '1':
+if os.environ.get('FORCE_QT_API') is not None:
     if 'PyQt5' in sys.modules:
         API = initial_api if initial_api in PYQT5_API else 'pyqt5'
     elif 'PySide2' in sys.modules:


JasonGilholme:
It looks like the python 2.7 tests fail due to an update to shiboken2 as PySide has been bumped from 5.12.3 -> 5.13 on June 27.

Previous run on master: https://circleci.com/gh/spyder-ide/qtpy/686
Latest run of master: https://circleci.com/gh/spyder-ide/qtpy/692

Release history of Shiboken2: https://pypi.org/project/shiboken2/#history

You'll need to address these changes at some point as anyone trying to use Qtpy on that version will experience an error when using the Qt3DAnimation module.

goanpeca:
Thanks for working on this @JasonGilholme.

I will open a new issues to track the problem -> see  https://github.com/spyder-ide/qtpy/issues/195

JasonGilholme:
No problem.

Has there ever been any discussion regarding joining efforts with the Qt.py
project?

I'm assuming they'll have to fix the same problem unless they have a design
that tolerates this sort of thing.

Regarding a comment you posted on my original issue, you could say that the
gain would be more development resources.

goanpeca:
>Has there ever been any discussion regarding joining efforts with the Qt.py
project?

Yes, they were but I guess the author of Qt.Py had different goals in mind and decided to create a new project with an unfortunate (similar) name.

JasonGilholme:
> > Has there ever been any discussion regarding joining efforts with the Qt.py
> > project?
> 
> Yes, they were but I guess the author of Qt.Py had different goals in mind and decided to create a new project with an unfortunate (similar) name.

Interesting, from your point of view, do you see a major difference in the goals of the two projects?

Don't get me started on the names.   It's bad enough having 2 libs let alone having basically the same names.

goanpeca:
>Interesting, from your point of view, do you see a major difference in the goals of the two projects?

To be honest I do not know what qt.py does since we do not use it. I think the author wanted to have a shim in a single file instead of the way we chose to do it with separate files etc...

Regarding the name, well it is really not our fault since this project was created  first.

JasonGilholme:
It's unfortunate that implementation detail has been confused with goals in
this regard.

For the VFX industry, it shouldn't be long until we could drop the need for
a shim layer completely.  This PR is still needed for us to continue to
enjoy using qtawesome though.

qtpy/__init__.py
```diff
@@ -114,15 +114,17 @@ class PythonQtWarning(Warning):
...
-    API = initial_api if initial_api in PYQT5_API else 'pyqt5'
-elif 'PySide2' in sys.modules:
-    API = initial_api if initial_api in PYSIDE2_API else 'pyside2'
-elif 'PyQt4' in sys.modules:
-    API = initial_api if initial_api in PYQT4_API else 'pyqt4'
-elif 'PySide' in sys.modules:
-    API = initial_api if initial_api in PYSIDE_API else 'pyside'
+# When `FORCE_QT_API` is set to "1", we disregard
+# any previously imported python bindings.
+if os.environ.get('FORCE_QT_API') == '1':
```
ccordoba12: Let's change this line to

    if os.environ.get('FORCE_QT_API') is not None:

to avoid relying on a predetermined value of `1`, which I find a bit odd.

In this case, if `FORCE_QT_API` is set to any value, then the code inside this `if` won't take effect.

JasonGilholme: Updated.

A value of 1 or 0 is pretty common for boolean environment variables, but it's not a technical feature of the shell.  A value like that can interface nicely in different languages; like javascript where `true == "1"` actually evaluates to True!

ccordoba12:
@JasonGilholme, please rebase in top of master to get the fix to our failing tests.

PR #204, opened 2019-12-07
PR: Add Python 3.9 compatibility for `collections.abc` module

### LINKED ISSUE #206: DeprecationWarning for getchildren
I'm running into a deprecation warning with PyQt5 5.14:

```
.../lib/python3.8/site-packages/qtpy/uic.py:184: DeprecationWarning: This method will be removed in future versions.  Use 'list(elem)' or iteration over elem instead.
     for custom_widget in custom_widgets.getchildren():
```

I guess the key here will be to figure out if it's safe to always iterate over ``custom_widgets`` directly or whether some older versions of PyQt and PySide need the explicit call to ``getchildren``.

tirkarthi:
* Importing ABC from collections module will raise `ImportError` in Python 3.9 and raises `DeprecationWarning` from 3.4 . Hence use `collections.abc` and also preserve compatibility with Python 2. Ref : https://github.com/python/cpython/pull/10596
* ~Use `list` instead of deprecated getchildren in Python 3. #206 #238~

### DIFF
--- a/qtpy/uic.py
+++ b/qtpy/uic.py
@@ -2,7 +2,7 @@ import os
 
 from . import PYSIDE, PYSIDE2, PYQT4, PYQT5
 from .QtWidgets import QComboBox
-from .pycompat import PY33
+from .py3compat import PY33
 
 
 if PYQT5:
--- a/qtpy/uic.py
+++ b/qtpy/uic.py
@@ -2,6 +2,7 @@ import os
 
 from . import PYSIDE, PYSIDE2, PYQT4, PYQT5
 from .QtWidgets import QComboBox
+from .pycompat import PY33
 
 
 if PYQT5:
@@ -180,8 +181,12 @@ else:
             return {}
 
         custom_widget_classes = {}
+        if PY33:
+            children = list(custom_widgets)
+        else:
+            children = custom_widgets.getchildren()
 
-        for custom_widget in custom_widgets.getchildren():
+        for custom_widget in children:
 
             cw_class = custom_widget.find('class').text
             cw_header = custom_widget.find('header').text
--- a/qtpy/uic.py
+++ b/qtpy/uic.py
@@ -2,7 +2,6 @@ import os
 
 from . import PYSIDE, PYSIDE2, PYQT4, PYQT5
 from .QtWidgets import QComboBox
-from .py3compat import PY33
 
 
 if PYQT5:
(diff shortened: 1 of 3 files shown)


tirkarthi:
I am not sure the build failures in some of the pipelines are related to my PR.

qtpy/uic.py
```diff
@@ -180,8 +181,12 @@ def _get_custom_widgets(ui_file):
             return {}
 
         custom_widget_classes = {}
+        if PY33:
```
goanpeca: We are not supporting this version of python, so why do we need this?

tirkarthi: Sorry for the late reply. PY33 is a variable to ensure the python version is 3.3 and above. As per the below deprecation it seems the methods were deprecated from 3.2. So I made sure to use `list(custom_widgets)` in Python 3.3+ and fallback to `getchildren` for versions below python 3.3 . Python 3.3 is listed in setup.py and this is the format used in most places for compatibility and hence my change.

> Methods getchildren() and getiterator() of classes ElementTree and Element in the ElementTree module have been removed. They were deprecated in Python 3.2. Use iter(x) or list(x) instead of x.getchildren() and x.iter() or list(x.iter()) instead of x.getiterator(). (Contributed by Serhiy Storchaka in bpo-36543.)

https://docs.python.org/3.9/whatsnew/3.9.html#removed

graingert:
@goanpeca it might be nicer to use `from six.moves.collections_abc import Callable`

it seems that it would be more idiomatic to use `six` everywhere in place of `qtpy.py3compat`

qtpy/uic.py
```diff
@@ -2,6 +2,7 @@
 
 from . import PYSIDE, PYSIDE2, PYQT4, PYQT5
 from .QtWidgets import QComboBox
+from .pycompat import PY33
```
dalthviz, line 5:
```suggestion
from .py3compat import PY33
```

dalthviz:
Hi @tirkarthi , after talking with @ccordoba12 I will do some changes here to wrap the work and move forward for the release (basically removing the PY33 logic and the line for the widgets iteration to then follow up the widgets change at #224). Hope this is ok with you @tirkarthi :)

dalthviz (approved):
Thank you @tirkarthi for your work in this one! :+1:

tirkarthi:
Thanks @dalthviz

Issue #210, opened 2020-03-18
Seems qtpy not support windows flags in QtCore.Qt ?


luojunyuan:
![image](https://user-images.githubusercontent.com/28323948/76920362-4796b200-6906-11ea-943a-8c40819ce217.png)

In the screenshot, pycharm can't recognize these flags but program goes well. 
use `from PyQt4.QtCore import Qt` is ok

And I have another question.
I'm using PyQt4.8 in my project. I use `QDeclarativeView` to create a qml windows. But in Qt5 it changed to `QQuickView` with another base class. But there is still a seem class named `QQuickWidget` in Qt5. Can qtpy make a binding to it?

Thanks!

goanpeca:
@luojunyuan feel free to make a PR with the changes you suggest.

luojunyuan:
oh, about the QtCore.Qt module I think it's a mistake made by PyCharm. Nothing wrong now

PR #211, opened 2020-04-08
Add QtQuick1 module to support QtDeclaretive in Qt4

luojunyuan:
#210 
QtDeclaretive has been replaced by the new QtQML and QtQuick modules in Qt5.  But the classes are not only named differently, but also used differently.
Qt5 used to have a Renamed QtDeclaretive called QtQuick1, but after Qt5.5 they removed that module entirely.

see https://doc.qt.io/qt-5/qtquick-porting-qt5.html#qtdeclarative-module-in-qt-5

### DIFF
--- a/qtpy/QtQuick1.py
+++ b/qtpy/QtQuick1.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+# -----------------------------------------------------------------------------
+# Copyright © 2009- The Spyder Development Team
+#
+# Licensed under the terms of the MIT License
+# (see LICENSE.txt for details)
+# -----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


goanpeca:
@luojunyuan is this ready?

luojunyuan:
> @luojunyuan is this ready?

what is ready mean?
I have run `python setup.py install` and use the module qtpy.QtQuick1 in my project now..

goanpeca:
@luojunyuan thanks for your contribution, however to accept it you will need to create a PR from a branch and not directly from master.

Please follow directions similar to https://github.com/spyder-ide/spyder/blob/master/CONTRIBUTING.md

Thanks!

PR #214, opened 2020-06-19
Handle QtCore.SignalInstance/pyqtBoundSignal

altendky:


goanpeca:
Hi @altendky thanks for the contribution. 

What is this exactly addressing 🙃 ?

altendky:
I was trying to type hint a function that takes a signal as a parameter.  The object you get when you `a_qobject.some_signal` isn't a `QtCore.Signal` but rather a `QtCore.SignalInstance` or `QtCore.pyqtBoundSignal` depending.

I'm not actually running mypy on this project yet but I think I'll need to do something over in the stubs as well based on PyCharm being a bit confused as is.  (or maybe I just haven't installed the third-party stubs yet and that'll help out).

goanpeca:
Ok understood. Anything that we could do on our side with MyPy to help with this?

altendky:
I really don't know.  I've only touched mypy a little bit at this point.  If I find something that seems like it would be best fixed here, I'll let you know.  :]  For now, just having a single way to refer to the 'signal instance' class would address my immediate need.  (well, I worked around it with a quick if in my own code so I'm not being held up).

goanpeca:
@altendky ok, so far this looks ok, will merge. A new release will be available in a couple of weeks while I do some cleanup of other PRs. Thanks again!

goanpeca:
We have some simple tests that check that things are imported. Could you add one?

altendky:
Yeah, I saw that but didn't see a lot so I skipped them.  Should be an easy add.  Maybe in a few hours here.

PR #208, opened 2020-02-05
PR: Move CI to Github Actions

goanpeca:
@ccordoba12 I added as much options as possible with github actions. 

Which versions should we test for python and which builds could we skip?

It is still pretty fast with the current setup.

PR #217, opened 2020-07-08
PR: Support mypy

### LINKED ISSUE #216: Support type hints and mypy
I am just starting into this but figured I'd get a ticket in sooner than later.  Hopefully supporting type hints requires pretty minimal effort in qtpy since it is just (primarily?) a pass through library.  I have had at least preliminary success by adding a [`py.typed` file](https://www.python.org/dev/peps/pep-0561/#packaging-type-information) and then using [mypy's `--always-true` and `--always-false`](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-always-true) to set `PYSIDE2` and `PYQT5` variables as they need to be for the library I am interested in.  I imagine giving qtpy a cli such that you could call `env/bin/mypy --target yourlib $(env/bin/qtpy mypyargs)` and the qtpy cli would generate the proper mypy arguments based on the regular library selection code.  For the sake of allowing mypy runs to really be static I would expect to also document everything to set and allow generation of args via the cli based on a forced library like `env/bin/qtpy mypyargs --use pyside2` or such.

Opinions?  What sort of target would I have to hit to have a PR for this ready for the next release?

altendky:
Fixes #216.

Draft for:
- [x] Convert to argparse for the CLI
- [x] Self review and retest since this has been sitting for awhile
- [ ] Wait for https://github.com/spyder-ide/qtpy/pull/208 to be merged so we have CI for these changes

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,24 @@ conda install qtpy
 ```
 
 
+### mypy
+
+A CLI is offered to help with usage of QtPy include the subcommands
+`qtpy mypy args` which generates command line arguments similar to the
+following which help guide mypy through which library QtPy would have used
+so that mypy can get the proper underlying type hints.
+
+```
+--always-false=PYQT4 --always-false=PYQT5 --always-false=PYSIDE --always-true=PYSIDE2
+```
+
+Use such as:
+
+```bash
+env/bin/mypy --package mypackage $(env/bin/qtpy mypy args)
+```
+
+
 ## Contributing
 
 Everyone is welcome to contribute!
--- a/README.md
+++ b/README.md
@@ -73,19 +73,20 @@ conda install qtpy
 
 ### mypy
 
-A CLI is offered to help with usage of QtPy including the subcommand
-`qtpy mypy-args` which generates command line arguments similar to the
-following which help guide mypy through which library QtPy would have used
-so that mypy can get the proper underlying type hints.
+A CLI is offered to help with usage of QtPy.  Presently, the only feature
+is to generate command line arguments for Mypy that will enable it to
+process the QtPy source files with the same API as QtPy itself would have
+selected.
 
 ```
 --always-false=PYQT4 --always-false=PYQT5 --always-false=PYSIDE --always-true=PYSIDE2
 ```
 
-Use such as:
+If using bash or similar, this can be integrated into the Mypy command line
+as follows.
 
-```bash
-env/bin/mypy --package mypackage $(env/bin/qtpy mypy-args)
+```console
+$ env/bin/mypy --package mypackage $(env/bin/qtpy mypy-args)
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ conda install qtpy
 ### mypy
 
 A CLI is offered to help with usage of QtPy including the subcommand
-`qtpy mypy args` which generates command line arguments similar to the
+`qtpy mypy-args` which generates command line arguments similar to the
 following which help guide mypy through which library QtPy would have used
 so that mypy can get the proper underlying type hints.
 
@@ -85,7 +85,7 @@ so that mypy can get the proper underlying type hints.
 Use such as:
 
 ```bash
-env/bin/mypy --package mypackage $(env/bin/qtpy mypy args)
+env/bin/mypy --package mypackage $(env/bin/qtpy mypy-args)
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ conda install qtpy
 
 ### mypy
 
-A CLI is offered to help with usage of QtPy include the subcommands
+A CLI is offered to help with usage of QtPy including the subcommand
 `qtpy mypy args` which generates command line arguments similar to the
 following which help guide mypy through which library QtPy would have used
 so that mypy can get the proper underlying type hints.
--- a/qtpy/__main__.py
+++ b/qtpy/__main__.py
@@ -1,0 +1,3 @@
+import qtpy.cli
+
+qtpy.cli.cli()
--- a/qtpy/__main__.py
+++ b/qtpy/__main__.py
@@ -1,3 +1,9 @@
 import qtpy.cli
 
-qtpy.cli.cli()
+
+def main():
+    return qtpy.cli.cli()
+
+
+if __name__ == "__main__":
+    main()
--- a/qtpy/cli.py
+++ b/qtpy/cli.py
@@ -17,7 +17,18 @@ def mypy():
 
 @mypy.command()
 def args():
-    """Generate command line arguments for using mypy with QtPy."""
+    """Generate command line arguments for using mypy with QtPy.
+
+    This will generate strings similar to the following which help guide mypy
+    through which library QtPy would have used so that mypy can get the proper
+    underlying type hints.
+
+        --always-false=PYQT4 --always-false=PYQT5 --always-false=PYSIDE --always-true=PYSIDE2
+
+    Use such as:
+
+        env/bin/mypy --package mypackage $(env/bin/qtpy mypy args)
+    """
 
     options = {False: '--always-false', True: '--always-true'}
 
--- a/qtpy/cli.py
+++ b/qtpy/cli.py
@@ -20,7 +20,7 @@ class RawDescriptionArgumentDefaultsHelpFormatter(
     pass
 
 
-def cli():
+def cli(args=sys.argv):
     """Features in support of development with QtPy."""
 
     parser = argparse.ArgumentParser(
@@ -46,9 +46,9 @@ def cli():
         description=inspect.getdoc(args),
         formatter_class=RawDescriptionArgumentDefaultsHelpFormatter,
     )
-    mypy_args_parser.set_defaults(func=args)
+    mypy_args_parser.set_defaults(func=mypy_args)
 
-    arguments = parser.parse_args()
+    arguments = parser.parse_args(args=args)
 
     reserved_parameters = {'func'}
     cleaned = {
@@ -66,7 +66,7 @@ def mypy():
     pass
 
 
-def args():
+def mypy_args():
     """Generate command line arguments for using mypy with QtPy.
 
     This will generate strings similar to the following which help guide mypy
--- a/qtpy/cli.py
+++ b/qtpy/cli.py
@@ -1,3 +1,14 @@
+# -*- coding: utf-8 -*-
+# -----------------------------------------------------------------------------
+# Copyright © 2009- The QtPy Contributors
+#
+# Released under the terms of the MIT License
+# (see LICENSE.txt for details)
+# -----------------------------------------------------------------------------
+
+"""Provide a CLI to allow configuring developer settings, including mypy."""
+
+# Standard library imports
 import argparse
 import inspect
 
--- a/qtpy/cli.py
+++ b/qtpy/cli.py
@@ -1,21 +1,60 @@
-import click
+import argparse
+import inspect
 
 
-@click.group()
-def cli():
-    """Features in support of development with QtPy."""
-
+class RawDescriptionArgumentDefaultsHelpFormatter(
+    argparse.RawDescriptionHelpFormatter,
+    argparse.ArgumentDefaultsHelpFormatter,
+):
     pass
 
 
-@cli.group()
+def cli():
+    """Features in support of development with QtPy."""
+
+    parser = argparse.ArgumentParser(
+        description=inspect.getdoc(cli),
+        formatter_class=RawDescriptionArgumentDefaultsHelpFormatter,
+    )
+
+    parser.set_defaults(func=parser.print_help)
+
+    cli_subparsers = parser.add_subparsers()
+
+    mypy_parser = cli_subparsers.add_parser(
+        name=mypy.__name__,
+        description=inspect.getdoc(mypy),
+        formatter_class=RawDescriptionArgumentDefaultsHelpFormatter,
+    )
+    mypy_parser.set_defaults(func=mypy)
+
+    mypy_subparsers = mypy_parser.add_subparsers()
+
+    mypy_args_parser = mypy_subparsers.add_parser(
+        name=args.__name__,
+        description=inspect.getdoc(args),
+        formatter_class=RawDescriptionArgumentDefaultsHelpFormatter,
+    )
+    mypy_args_parser.set_defaults(func=args)
+
+    arguments = parser.parse_args()
+
+    reserved_parameters = {'func'}
+    cleaned = {
+        k: v
+        for k, v in vars(args).items()
+        if k not in reserved_parameters
+    }
+
+    arguments.func(**cleaned)
+
+
 def mypy():
     """Features in support of running mypy"""
 
     pass
 
 
-@mypy.command()
 def args():
     """Generate command line arguments for using mypy with QtPy.
 
--- a/qtpy/cli.py
+++ b/qtpy/cli.py
@@ -3,16 +3,22 @@ import click
 
 @click.group()
 def cli():
+    """Features in support of development with QtPy."""
+
     pass
 
 
 @cli.group()
 def mypy():
+    """Features in support of running mypy"""
+
     pass
 
 
 @mypy.command()
 def args():
+    """Generate command line arguments for using mypy with QtPy."""
+
     options = {False: '--always-false', True: '--always-true'}
 
     import qtpy
--- a/qtpy/cli.py
+++ b/qtpy/cli.py
@@ -10,7 +10,8 @@
 
 # Standard library imports
 import argparse
-import inspect
+import sys
+import textwrap
 
 
 class RawDescriptionArgumentDefaultsHelpFormatter(
@@ -20,11 +21,9 @@ class RawDescriptionArgumentDefaultsHelpFormatter(
     pass
 
 
-def cli(args=sys.argv):
-    """Features in support of development with QtPy."""
-
+def cli(args=sys.argv[1:]):
     parser = argparse.ArgumentParser(
-        description=inspect.getdoc(cli),
+        description="Features in support of development with QtPy.",
         formatter_class=RawDescriptionArgumentDefaultsHelpFormatter,
     )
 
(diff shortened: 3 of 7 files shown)


altendky:
I'm using this branch over in https://github.com/altendky/qtrio/pull/83 if you are interested in seeing it running.  It would be nice to add mypy runs in QtPy's CI as well.  Is there any expected timeline for #208?  Perhaps there is something I could help with to get it done?  I'm just figuring I may as well add mypy to the new CI rather than the existing/old.  Or, perhaps mypy checks against QtPy should just be added separately.

altendky:
Welp, there's piles of work in both PySide2 and PyQt5 hints but so far this seems to be ok.  Any concerns here?  Or anything I can do to help?

CAM-Gerlach (commented):
Sorry for the very long delay; unfortunately, the main developer who was responsible for this left and resources were not re-allocated appropriately to maintain this critical package. While I don't actually use it myself (outside of as a Spyder developer, and I'm not that active anymore), given its clearly still well used by the community, I'll do my best to try to put out fires and get things moving along again, with your help. Thanks so much @altendky for your dedication! It really means a lot!

Overall, this seems like a fine change, if its needed to make things easier for mypy. However, instead of adding a new dependency on Click (and its peculiar syntax) just for a relatively trivial CLI, itself only for developer use with mypy, why not just use the standard-library `argparse`? Other than that, I'm :+1: , once I get the CIs working again in PR #208 . Thanks!

altendky:
Why Click?  Because it is \_so\_ much nicer.  `:]`  And, as you say, it doesn't need to be (and isn't) a runtime dependency.  Some programs are mostly about doing one thing with a few options, but a QtPy CLI seems more likely to be made up of miscellaneous helpers for different activities.  This lends itself very much to subcommands which aren't the most pleasant thing in argparse.

But sure, I'll go ahead and switch this over to argparse.  I've set this to draft and added notes in the OP listing what activities I should do prior to releasing this for review again.

@CAM-Gerlach, thanks for finding some time here.  If you want to tag me in on other tasks, let me know and I'll do whatever I can.

qtpy/cli.py
```diff
@@ -0,0 +1,81 @@
...
+
+
+class RawDescriptionArgumentDefaultsHelpFormatter(
+    argparse.RawDescriptionHelpFormatter,
+    argparse.ArgumentDefaultsHelpFormatter,
+):
+    pass
+
+
+def cli():
```
CAM-Gerlach, line 12:
```suggestion
def cli(cli_args=None):
```

Wouldn't hurt to have this, in case we want to pass in arguments programmatically

qtpy/cli.py
```diff
@@ -0,0 +1,81 @@
...
+    mypy_subparsers = mypy_parser.add_subparsers()
+
+    mypy_args_parser = mypy_subparsers.add_parser(
+        name=args.__name__,
+        description=inspect.getdoc(args),
+        formatter_class=RawDescriptionArgumentDefaultsHelpFormatter,
+    )
+    mypy_args_parser.set_defaults(func=args)
+
+    arguments = parser.parse_args()
```
CAM-Gerlach, line 40:
```suggestion
    arguments = parser.parse_args(args=cli_args)
```

As above

qtpy/cli.py
```diff
@@ -0,0 +1,81 @@
...
+    options = {False: '--always-false', True: '--always-true'}
+
+    import qtpy
+
+    apis_active = {name: qtpy.API.lower() == name.lower() for name in qtpy.APIS}
+    print(' '.join(
+        f'{options[is_active]}={name.upper()}'
+        for name, is_active
+        in apis_active.items()
+    ))
```
CAM-Gerlach, line 81:
```suggestion
    ))


def main():
    cli()


if __name__ == "__main__":
    main()
```

A bit of boilerplate to make it a little easier to run as a script in case the user doesn't actually have QtPy installed, but rather just checked out locally

qtpy/__main__.py
```diff
@@ -0,0 +1,3 @@
+import qtpy.cli
+
+qtpy.cli.cli()
```
CAM-Gerlach, line 3:
```suggestion

def main():
    qtpy.cli.cli()


if __name__ == "__main__":
    main()
```

I usually like to do things this way, as recommended in the setuptools docs, so I can point my setuptools entrypoint here so it doesn't break in case the cli file is moved, the function is renamed, or something else similar is changed.

setup.py
```diff
@@ -47,5 +47,7 @@
         'Programming Language :: Python :: 3',
         'Programming Language :: Python :: 3.3',
         'Programming Language :: Python :: 3.4',
-        'Programming Language :: Python :: 3.5']
+        'Programming Language :: Python :: 3.5'],
+    entry_points={'console_scripts': 'qtpy = qtpy.cli:cli [cli]'},
```
CAM-Gerlach, line 51:
```suggestion
    entry_points={'console_scripts': 'qtpy = qtpy.__main__:main'},
```

To ensure our main entry point doesn't need to be updated if it changes

CAM-Gerlach (changes requested):
All sounds good! Just some minor nitpicks below. Hopefully we can get PR #208 merged soon, I just want to give @ccordoba12 a chance to look over it before going ahead and merging.

CAM-Gerlach:
Yeah, I really like the idea of Click; I'm just used to argparse and its tended to be a decent fit for the stuff I build, which does have a lot of subcommands (albeit not sub-sub-commands) and I've developed patterns (albiet boilerplate-heavy) to work around most of its limitations. In this case, I didn't know you'd actually re-implement a bunch of Click's features within argparse for this relatively simple usecase, rather than just doing something simple; sorry for all the effort or I'd have just said to keep Click, ha!

As for help, we'd welcome whatever you would like to contribute! Our main priority right now is getting the backlog of PRs reviewed, tested and merged as soon as practical, so we can release 1.10 with the accumulated changes. Once that's done, it will be much easier to work on things (esp things like adding Mypy annotations or type stubs, if you're thinking of that) without merge conflicts or things breaking. We'd also welcome help addressing open issues that aren't already covered by PRs, or anything else you find, as well as giving feedback and testing on them.

Then, what I'm thinking might be a good tentative plan beyond that is to clean house and drop Py2 and Qt4, which we're no longer able to properly test due to package availability, add Qt6 support with PR #233 and release that as 2.0,  but that's just my internal thinking, if @ccordoba12 and @stonebig agree. Maybe also add pre-commit to automate a good deal of the tedium of review and maintenance.

Again, many thanks!

altendky:
Can we just go back to Click?  `:]`  It's massively simpler and doesn't involve forgoing features nor recreating them.  Though, it only took a few minutes for this.  I do also have my own argparse patterns I've made, which these came from.  Mostly only used in my own env-bootstrapper that of course generally ends up installing Click so everything else can use it.

https://github.com/altendky/boots/blob/9d399eed1fb7448dd3be7e707a787202f1755918/boots.py#L1018-L1171

It's not like it is a runtime dependency.  It also is a pretty miniscule dependency compared with PyQt or PySide.  It's even pure Python.

Anyways, I'll go through all your notes.  I generally also have reasons for doing things the way I do.  But, the goal here is to build some momentum with QtPy.  I was getting darn close to forking it...  I'm glad there's renewed hope to keep the work here.  `:]`

For the hints, I think that mostly just passes through to PyQt and PySide.  Other than the helpful hints for Mypy about what API is in use, as added here with the CLI.

stonebig:
@CAM-Gerlach I'm so happy you move Qtpy forward to Qt6, so I agree with anything.

CAM-Gerlach:
@stonebig Don't thank me haha, it was you and @jschueller that have done all the work on that front! I'm just here to help with maintenance, review, getting PRs moved forward and hopefully cutting releases. I was just about to ask on PR #225 , but I'd welcome your feedback on the tentative plan above; I'll comment on #225 so its more visible to everyone participating there. Thanks so much for your hard work!

CAM-Gerlach:
> Can we just go back to Click? :] It's massively simpler and doesn't involve forgoing features nor recreating them.

Actually I was thinking about that, given the considerable code complexity that using argparse entails given the (admittedly rather Click-centric) design choices made. However, upon further consideration, there are a few reasons this doesn't end up being a net win for simplicity, especially from a maintainer perspective:

* Conda (through which many if not most of our users get QtPy) does not support optional dependencies; there are workarounds but they add significant effort and complexity for maintainers, users or both; see conda/conda#7502
* It will require some feedstock changes to get the conda builds working, at a minimum adding click as a test-only dependency (I think that's possible to do), on top of whatever approach we take to resolve the above. I don't have direct access there, so it could delay a conda-forge release.
* It isn't used in any other Spyder-related project (nor the broader scientific Python ecosystem) and none of us are really familiar with it, making things more difficult in the event of a bug, breaking change or new adding a new command.

If QtPy had a greater need for a general-purpose CLI, these (relatively modest) costs could be outweighed by its considerable benefits for such a use case, but as it stands, I just don't see how its worthwhile for a single, simple, developer-only command.

However, much of the complexity (most notably having multiple layers of subcommands in subcommands) appears to be unnecessary or for the very simple use case presented here, of a single command returning a single string for consumption by another tool, intended for a small subset of developers. Simplifying it actually ends up being the same total length (0 net line change of code + tests) as the original version using Click, at minimal loss of meaningful functionality. See [my repo](https://github.com/cam-gerlach/qtpy/tree/mypy) for an implementation; feel free to pull that in to yours here or I can commit it directly to your branch, if you want. Also, I tweaked your `args()` function to actually return its args so it can be called programmatically, and renamed it more descriptively as a verb following standard PEP 8 convention. I've also fixed a bug in the tests on Windows in a separate commit.

> I generally also have reasons for doing things the way I do.

Yeah, my suggestions were just minor stuff: adding our standard file header, fixing a few grammar issues in the text, capitalizing a constant, and tweaking a few things about the entrypoint setup. I am curious for your input particularly on the last one—I've wondered if there's a reason/benefit to not pointing to `__main__` for the main top-level entrypoint, that would outweigh the downsides mentioned.

Also, BTW keeping an eye on your ciborg project—would be really nice to have a CI abstraction layer; now that Conda-Smithy has Github Actions support I've been considering that, but as its pretty specialized a general tool would be much nicer.

altendky:
The nested subcommands were intended to lay out an organized interface that wouldn't require breaking changes if future tools are added.  If hazarding such changes is preferred, certainly a single subcommand such as `qtpy mypy-args` could be used instead.

I'll provide feedback on the rest.

It's sad that Conda, used to make it easier to get stuff, makes it harder to get stuff.  `:[`

qtpy/__main__.py
altendky: I don't see the major upside as compared with encouraging the hazardous behavior of making files both directly runnable and importable.

setup.py
altendky: This demands a whole turn around that includes some icky stuff I'd rather just avoid.  Not worth maybe having to update an entry point someday, maybe.

CAM-Gerlach:
Good news, I went ahead and merged PR #208, so if you rebase you should get the tests running, and this should hopefully be close to ready.

> The nested subcommands were intended to lay out an organized interface that wouldn't require breaking changes if future tools are added. If hazarding such changes is preferred, certainly a single subcommand such as qtpy mypy-args could be used instead.

Its not a bad approach, but with multiple layers of nested subcommands being a relatively uncommon design pattern for at least this ecosystem, increasing API, code and test complexity, and adding a new dev/test dependency with, all for a single specialized command seems a little overengineered to me (something I'm very much guilty of myself often). I'm also a little confused where the breaking changes come in; future tools could simply be added in separate subcommands, if the functionality cannot be added to the existing in a backward-compatible way.

> I'll provide feedback on the rest.

Okay, thanks. I'll take a look momentarily.

> It's sad that Conda, used to make it easier to get stuff, makes it harder to get stuff. :[

Yeah, unfortunately the design decisions they've taken on that aspect have made it more difficult than it should be to implement optional dependencies in an easily maintainable manner. On the other hand, at least their build automation and tooling makes things that would be very painful and brittle with `pip`/`setuptools`/`wheel` relatively straightforward.

README.md
```diff
@@ -71,6 +71,24 @@ conda install qtpy
 ```
 
 
+### mypy
+
+A CLI is offered to help with usage of QtPy including the subcommand
+`qtpy mypy args` which generates command line arguments similar to the
+following which help guide mypy through which library QtPy would have used
+so that mypy can get the proper underlying type hints.
```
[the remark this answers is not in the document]
CAM-Gerlach: Sorry if I've impaired the meaning, I was trying to break up a long run-on sentence and not have the "similar to the following" be disconnected from said "following". Your new revision looks much better than either, so we can go from there, thanks.

README.md
```diff
@@ -71,6 +71,24 @@ conda install qtpy
...
+
+```
+--always-false=PYQT4 --always-false=PYQT5 --always-false=PYSIDE --always-true=PYSIDE2
+```
+
+Use such as:
```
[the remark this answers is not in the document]
CAM-Gerlach: I guess my intention was that its better to be explicit rather than implicit about what users are expected to replace, if the example is intended to be of practical value. However, given your point, perhaps its simpler to just specify the default invocations of the `mypy` and `qtpy` installed in the current environment (i.e. `mypy` and `qtpy`), which users can replace themselves with someone more specific if their situation requires it? Or is there some particular reason that `mypy` needs to be invoked with the full path?

qtpy/tests/test_cli.py
```diff
@@ -0,0 +1,45 @@
...
+
+import subprocess
+import sys
+
+import pytest
+
+import qtpy
+
+
+subcommands = [
```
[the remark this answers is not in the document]
CAM-Gerlach: Given Python has no runtime enforcement of constants staying constants, particularly for mutable objects like this list here which are effectively global state, it is particularly important to clearly distinguish them from regular variables. `UPPER_CASE` is the standard convention for global constants per [PEP 8](https://www.python.org/dev/peps/pep-0008/#constants), [Google style](https://google.github.io/styleguide/pyguide.html#s3.16-naming) and pretty much every other Python style guide I'm aware of, and is widely adopted in first and third-party packages, including QtPy.

qtpy/cli.py
CAM-Gerlach: Good points, agreed this isn't necessary or desirable.

> And why the indirection through main()?

Just convention, and probably a foolish one, at least in this case where it doesn't do anything special.

> Would want a sys.exit()?

Not sure the need for `sys.exit()`, when it would exist anyway with `0` upon completion?

qtpy/__main__.py
CAM-Gerlach: I'm afraid I don't follow the reasoning here. Could you explain why this behavior is more hazardous in this context of a trivial module following a well-known common pattern, relative to the practical risk of having separate `console_scripts` and runpy entrypoints that may behave differently or get out of sync? The `if __name__ == "__main__"` idiom is very common (the [question](https://stackoverflow.com/questions/419163/what-does-if-name-main-do) explaining how and why to use it is the second most popular tagged `python` on all of SO), especially in the scientific Python ecosystem, and is recommended in the [official Python docs](https://docs.python.org/3/library/__main__.html) and countless other sources.

Further, this particular pattern is recommended a [substantial](https://chriswarrick.com/blog/2014/09/15/python-apps-the-right-way-entry_points-and-scripts/) [number](https://stackoverflow.com/questions/27784271/how-can-i-use-setuptools-to-generate-a-console-scripts-entry-point-which-calls) of [places](https://stackoverflow.com/questions/44977227/how-to-configure-main-py-init-py-and-setup-py-for-a-basic-package), none of which describe any such issues with it.

Of course, there is the general principle of separation of concerns, and thus that entry points and package functionality should be separate (which this still fully ensures, as this module contains no non-trivial functionality beyond that necessary to serve as a single-source entrypoint for both `console_scripts` and `runpy`). 

Beyond that, the only directly relevant post I found explicitly advocating against any use of `if __name__ == __main__` was [this personal blog](https://moshez.wordpress.com/2016/06/07/__name__-__main__-considered-harmful/), the one comment to which was a rebuttal by a Python core dev, and the (rather esoteric) potential harms mentioned don't apply to a module containing no non-trivial functionality beyond re-directing to the entry point. Some of these are at least theoretically possible if `qtpy,cli` were made runnable, but that's not what is being proposed here, and following your suggestions above, I rejected that. Therefore, I cannot see what the practical harm is, relative to the practical benefit (and avoidance of much more likely harm) of single-sourcing the entry piont.

setup.py
CAM-Gerlach: I'm a little confused with how changing a few characters here and adding a couple lines of standard boilerplate to `__main__,py`, both of which I made for you as suggestions, "demands a whole turn around" in your approach. Furthermore, as explained above, could you explain what practical harms this "icky stuff" has in a trivial module containing no actual logic that outweigh the non-trivial risk and cost of maintaining two parallel entrypoints?

CAM-Gerlach:
Hey, so unfortunately, by the time I was done reviewing and responding to the comments here, its 4am now and I need to eat and sleep, so I'll have to finish the actual review tomorrow, sorry. However, I did notice a number of things missing that were fixed in the linked branch on my fork, including fixing the tests on Windows, making `mypy_args` actually return the args so they could be retrieved programmatically (without hacks) and then simply wrapping that in a print lambda in `cli()`, resolving grammar errors in the docstring, handling the default `argv` properly, and making the function names verbs. I'm also unsure why the docstrings were removed; the previous method of parsing them directly from the called function was no more verbose and allowed them to be used both as docstrings and CLI help.

altendky:
Just fix everything as you wish.

dalthviz:
Hi @MinmoTech thank you for your interest! No for the moment but I think this is something we want to finish for sure in the future. Maybe after we release [v2.1.0](https://github.com/spyder-ide/qtpy/milestone/34) we could take a look at this again and finish it. What do you think @CAM-Gerlach @ccordoba12 ?

Also, just in case, is okay for you @altendky if we (probably @CAM-Gerlach or myself) finish this? Quite late to say but thank you so much for all the patience and time you spent working on this!

ccordoba12:
Sure, I think this is useful.

CAM-Gerlach:
Sounds good to me, I have [a branch on my fork](https://github.com/CAM-Gerlach/qtpy/tree/mypy) implementing the requested changes (since @altendky gave the go-ahead above for us finishing it), but unfortunately life stuff intervened and I had to take a break from GitHub for a while. I'd been meaning to come back to finally finish it, I just hadn't quite gotten to it yet.

dalthviz:
> If you agree, I can push that here with the merge conflicts fixed and any other updates, and we can proceed from there. Or, I can open a new PR, your call.

I guess then merging/pushing your changes here should be okay @CAM-Gerlach , thanks!

CAM-Gerlach:
I ended up just opening a new PR, #337 , with the changes up until now squashed to a single commit to make rebasing easier, and then updated it with the various QtPy 2.x changes, fixed and improved the tests, simplified the CLI logic, refined the text/readme/help and added and tested a `--version` arg.

ccordoba12:
Ok, I'm going to close this PR then. Thanks for your help with this @altendky!

PR #221, opened 2020-11-03
More informative error message

cbrnr:
My project MNELAB depends on `qtpy` because I would like to support both `PyQt5` and `PySide2`. However, if no bindings are installed, the displayed error message could be more informative and tell users what to do (see e.g. the problem reported in https://github.com/cbrnr/mnelab/issues/182). I've added explicit instructions how to install either `PyQt5` or `PySide2`.

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -213,8 +213,7 @@ if API in PYSIDE_API:
         PYSIDE = True
     except ImportError:
         raise PythonQtError('No Qt bindings could be found, please install '
-                            'either PyQt5 or PySide2 (e.g. "pip install PyQt5"'
-                            ' or "pip install PySide2").')
+                            'either PyQt5 or PySide2.')
 
 # If a correct API name is passed to QT_API and it could not be found,
 # switches to another and informs through the warning
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -212,7 +212,9 @@ if API in PYSIDE_API:
         PYQT5 = PYSIDE2 = False
         PYSIDE = True
     except ImportError:
-        raise PythonQtError('No Qt bindings could be found')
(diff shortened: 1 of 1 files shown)


cbrnr:
@goanpeca @altendky do you have time to review this minor change?

altendky:
I have no authority here but...  I would tend to avoid explicit commands.  If people don't know to run pip to install stuff then they probably won't be remembering to use `venv/bin/pip` (you do correctly do everything in an env, right? `:]`) or `--user` or `apt install python3-pyqt5` or...  etc.

cbrnr:
I would assume that people know how to `pip install` stuff, but currently the problem is that they don't know that a package is missing.

altendky:
In which case they don't need to be given a command.  I understand the interest in saying more than just 'not found'.  Maybe something like `This is usually caused by not having installed a compatible Qt wrapper`.  Maybe even include the list of compatible wrappers.  (#217 happens to add a more explicit list (`dict`) of the supported wrappers)  Or maybe link to https://github.com/spyder-ide/qtpy#requirements?  Though then you can have a disconnect between the site and the installed version.

As a side note, I presently add `extras_requires` for the qt wrappers so people can install my packages like `package[pyside2]` to allow my package to specify whatever version restrictions etc it has for pyside2.  Or `package[pyqt5]` for that option.

cbrnr:
I agree, a command is not necessary, but I think the message should mention the wrappers explicitly to get people on the right track. Also, I would restrict the wrappers to `PySide2` and `PyQt5`, because `PyQt4` is old and deprecated.

Thanks for mentioning the `extras_require` thing, I haven't thought about that &ndash; that's a very good idea!

cbrnr:
OK, I've removed the part about `pip` from the message.

PR #223, opened 2020-11-18
Add missing trove classifiers with setup-py-upgrade and setup-cfg-fmt

graingert:
I noticed there were no supported python versions (py3.6, py3.7 and py3.8) in the pypi metadata

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,2 +1,37 @@
+[metadata]
+name = QtPy
+keywords = qt PyQt4 PyQt5 PySide
+url = https://github.com/spyder-ide/qtpy
+license = MIT
+author = Colin Duquesnoy, The Spyder Development Team
+author_email = goanpeca@gmail.com
+maintainer = Gonzalo Peña-Castellanos
+maintainer_email = goanpeca@gmail.com
+description = Provides an abstraction layer on top of the various Qt bindings (PyQt5, PyQt4 and PySide) and additional custom QWidgets.
+long_description = file: README.md
+long_description_content_type = text/markdown
+classifiers =
+    Development Status :: 5 - Production/Stable
+    Environment :: X11 Applications :: Qt
+    Environment :: Win32 (MS Windows)
+    Intended Audience :: Developers
+    License :: OSI Approved :: MIT License
+    Operating System :: OS Independent
+    Programming Language :: Python :: 2
+    Programming Language :: Python :: 2.7
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.3
+    Programming Language :: Python :: 3.4
+    Programming Language :: Python :: 3.5
+
+[options]
+packages = find:
+
+[options.packages.find]
+exclude =
+    contrib
+    docs
+    tests*
+
 [bdist_wheel]
-universal=1
+universal = 1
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -1,2 +1,42 @@
...
+    Operating System :: OS Independent
+    Programming Language :: Python :: 2
+    Programming Language :: Python :: 2.7
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.3
+    Programming Language :: Python :: 3.4
+    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
```
graingert, line 28: these are the only supported python versions

graingert: 3.9 support is in https://github.com/spyder-ide/qtpy/pull/204

CAM-Gerlach: Do we really still support Python 3.3 and 3.4? They're long-obsolete, Spyder itself dropped them ages ago.

setup.py
```diff
@@ -6,46 +6,13 @@
...
-        'License :: OSI Approved :: MIT License',
-        'Operating System :: OS Independent',
-        'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 2.7',
-        'Programming Language :: Python :: 3',
-        'Programming Language :: Python :: 3.3',
-        'Programming Language :: Python :: 3.4',
-        'Programming Language :: Python :: 3.5']
-)
+setup(version=version_ns['__version__'])
```
CAM-Gerlach, line 18: The version attr can be specified declaratively, as e.g. I do [here](https://github.com/project-mjolnir/brokkr/commit/775701c5b876a1b831094b4522d23b98da9abb93), so this last bit of dynamic code in the setup.py can be eliminated

setup.cfg
```diff
@@ -1,2 +1,42 @@
...
+description = Provides an abstraction layer on top of the various Qt bindings (PyQt5, PyQt4 and PySide) and additional custom QWidgets.
+long_description = file: README.md
+long_description_content_type = text/markdown
+url = https://github.com/spyder-ide/qtpy
+author = Colin Duquesnoy, The Spyder Development Team
+author_email = goanpeca@gmail.com
+maintainer = Gonzalo Peña-Castellanos
+maintainer_email = goanpeca@gmail.com
+license = MIT
+license_file = LICENSE.txt
```
CAM-Gerlach, line 12: `license_file` is deprecated and `license_files` should be used instead, and in any case `LICENSE.txt` will be included automatically, so this can just be removed.

```suggestion
```

CAM-Gerlach (commented):
Thanks for this, and sorry for the long delay! Just a few minor comments.

setup.cfg
CAM-Gerlach: This should be resolved in PR #208 ; you'll need to re-run your tooling once that's merged since it changes a bunch of the metadata

ccordoba12:
Thanks for your help with this @graingert. We already updated our classifiers in 1.10.0 and don't accustom to declare metadata in `setup.cfg` in our projects.

So we decided to close this one, but thanks for your interest in the improving this project.

graingert:
Is there a reason you prefer the legacy metadata format? @ccordoba12

ccordoba12:
It's just a matter of being used to do that in all our projects. So we prefer to keep things like that for now.

PR #227, opened 2021-01-14
Rename QtDatavisualization.py to use uppercase v

antlarr:
Both `PyQt5` and `PySide2` call it QtDatavisualization (with an uppercase v) and the `tests/test_qtdatavisualization.py` file even does `pytest.importorskip("qtpy.QtDataVisualization")` which means the test is currently always skipped since the module didn't exist.
    
Another solution would be to fix the test but for homogeneity with PyQt5 and PySide2 I think it's better to rename QtDatavisualization.py to QtDataVisualization.py .


antlarr:
I'm not sure why these commits were added with a different hash. I'll close and try to create submit again

PR #228, opened 2021-01-14
PR: Rename QtDatavisualization to use uppercase v

antlarr:
Both `PyQt5` and `PySide2` call the module QtDatavisualization (with an uppercase v) and the `tests/test_qtdatavisualization.py` file even does `pytest.importorskip("qtpy.QtDataVisualization")` which means the test is currently always skipped since the module didn't exist.

Another solution would be to fix the test but for homogeneity with PyQt5 and PySide2 I think it's better to rename QtDatavisualization.py to QtDataVisualization.py .


bnavigator:
Should QtPy provide an alias set in `__init__.py` for backwards compatibility?

CAM-Gerlach (commented):
Makes sense to me, esp. with the backward compat. However, I am kinda unclear on the PR description; you say

> Both `PyQt5` and `PySide2` call the module QtDatavisualization (with an uppercase v)

but while you say they call the module "with an uppercase v", the actual text written there as to what they call the module has a _lowercase_ v, so I'm a bit confused. Other than that, LGTM FWIW once we get the CIs running again (not sure what happened with them).

antlarr-suse:
Oops, that's a very unfortunate typo. I should have written `QtDataVisualization` indeed.

antlarr:
Oops, that's a very unfortunate typo. I should have written QtDataVisualization indeed.

CAM-Gerlach (commented):
Thanks; FWIW, LGTM. You might need to repush (e.g. `git commit --amend --no-edit && git push -f`) to trigger the checks once PR #208 is merged, which should hopefully happen shortly.

dalthviz:
Closing (and reopening) to check if a CI rerun is triggered

dalthviz (approved):
Thank you @antlarr ! LGTM :+1:

PR #235, opened 2021-04-15
deprecate distutils.LooseVersion

stonebig:


qtpy/__init__.py
```diff
@@ -62,7 +62,11 @@
 
 """
 
-from distutils.version import LooseVersion
+try:  # disutils is being deprecated
+    import pkg_resources
```
jschueller, line 66: this import is useless

qtpy/__init__.py
```diff
@@ -62,7 +62,11 @@
...
+    from pkg_resources import parse_version as LooseVersion
+except:
```
jschueller, line 68: except ImportError:

jschueller:
I'd rather use the new name with the old symbol than the new symbol with the old name: use parse_version instead.

Also it seems pkg_resources.parse_version is available since at least 2.7, so there maybe even be no need for a fallback, just using parse_version

qtpy/__init__.py
stonebig: I kept a backup plan, as I'm not sure on what very old Python QtPy may be used.

jschueller: ok, but it still useless

ccordoba12:
> Also, just in case, what do you think @ccordoba12 ?

I also agree with using `packaging` because `pkg_resources` is deprecated. My only question is: is `packaging` a `setuptools` vendored dependency? If so, what's the earliest version of `setuptools` we can rely on that includes it?

CAM-Gerlach:
@ccordoba12 It is a setuptools-vendored dep in recent versions, but we cannot rely on the Setuptools-vendored version as it is a private, internal-only implementation detail of setuptools, varies arbitrarily between setuptools versions (some may be out of date, while some may contain breaking changes), has setuptools-specific modifications, and can change or be dropped at any time. As `packaging` is a core officially-supported component of the PyPA ecosystem, is depended upon by many other major projects and is a lightweight pure-Python package that's only 40 kB, it seems sensible to simply add as a proper dependency under `install_requires` and the Conda recipe equivalent.

ccordoba12:
> it seems sensible to simply add as a proper dependency under install_requires and the Conda recipe equivalent.

Ok, I'm fine with that.

PR #230, opened 2021-02-08
PR: Fix imported modules logic if 'FORCE_QT_API' is empty

hiaselhans:
As mentioned in #222 i stumbled across this.

before:
Bildschirmfoto vom 2021-02-08 12-44-08

after (expected behaviour):
Bildschirmfoto vom 2021-02-08 12-47-17

It seems obvious to me and i spare further explination.

PR #237, opened 2021-05-03
Improve __init__.py.

UmbralReaper:
This is my attempt to clean up the login in `__init__.py`. It fixes #230, and also performs a general cleanup of the file. I'll also attempt to merge some of the functionality from #156 into this.

The biggest change currently is that if the user specifies a specific `QT_API`, and it is not found, a `PythonQtInvalidBinding`  `RuntimeError` wil be raised instead of a `RuntimeWarning`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,7 @@ jobs:
   test:
     name: Test ${{ matrix.os }} Python ${{ matrix.python-version }} conda=${{ matrix.use-conda }}
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 10
     defaults:
       run:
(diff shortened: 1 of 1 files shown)


tlambert03:
I agree, `__init__` needs a lot of love.  There was some additional possible cleanups in #247 that you might also have a look through and potentially incorporate here.  in particular ... I think the `If API ==` chain could be made cleaner.  see [here](https://github.com/tlambert-forks/qtpy/blob/49db3eb733cc90194dcc62d56fa7be8a436cab04/qtpy/__init__.py#L43-L71) for example

UmbralReaper:
Most of the changes I wanted to merge have already been merged during the process of removing PyQt4 support.

Issue #238, opened 2021-05-29
PySide2 and Python3.9: xml.etree.ElementTree.Element' object has no attribute 'getchildren


AndrewAmmerlaan:
With `QT_API="pyside2"`, QtPy-1.9.0 and python3.9, I get the following test failure:
```
=============================================================================================== FAILURES ================================================================================================
_______________________________________________________________________________________ test_load_ui_custom_auto ________________________________________________________________________________________

tmpdir = local('/var/tmp/portage/dev-python/QtPy-1.9.0-r4/temp/pytest-of-portage/pytest-5/test_load_ui_custom_auto0')

@pytest.mark.skipif((PYSIDE2 and os.environ.get('CI', None) is not None),
reason="It segfaults in our CIs with PYSIDE2")
def test_load_ui_custom_auto(tmpdir):
"""
Test that we can load a .ui file with custom widgets without having to
explicitly specify a dictionary of custom widgets, even in the case of
PySide.
"""

app = get_qapp()

with enabled_qcombobox_subclass(tmpdir):
from qcombobox_subclass import _QComboBoxSubclass
>           ui = loadUi(os.path.join(os.path.dirname(__file__), 'test_custom.ui'))

_QComboBoxSubclass = <class 'qcombobox_subclass._QComboBoxSubclass'>
app        = <PySide2.QtWidgets.QApplication(0x559c6cc2d810) at 0x7f2e425790c0>
tmpdir     = local('/var/tmp/portage/dev-python/QtPy-1.9.0-r4/temp/pytest-of-portage/pytest-5/test_load_ui_custom_auto0')

qtpy/tests/test_uic.py:71: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
qtpy/uic.py:219: in loadUi
customWidgets = _get_custom_widgets(uifile)
baseinstance = None
uifile     = '/var/tmp/portage/dev-python/QtPy-1.9.0-r4/work/QtPy-1.9.0/qtpy/tests/test_custom.ui'
workingDirectory = None
[...]

PR #224, opened 2020-11-19
PR: Support python 3.9 `custom_widgets` iteration

irrcombat:
Support python 3.9

Fixes #206 #238

PR #239, opened 2021-06-01
qtpy/uic.py: Make compatible with python3.9

AndrewAmmerlaan:
This fixes https://github.com/spyder-ide/qtpy/issues/238

Signed-off-by: Andrew Ammerlaan <andrewammerlaan@gentoo.org>

### DIFF
--- a/qtpy/uic.py
+++ b/qtpy/uic.py
@@ -181,7 +181,7 @@ else:
 
         custom_widget_classes = {}
 
-        for custom_widget in custom_widgets.getchildren():
+        for custom_widget in list(custom_widgets):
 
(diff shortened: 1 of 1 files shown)


hiaselhans:
seems to be a duplicate of #224 

still: good to make some noise for @ccordoba12 @andfoy

;)

dalthviz:
Hi @irrcombat thank you for you contribution and sorry for the late feedback! However as @hiaselhans mentions there is already #224 and checking the tracker even an older PR (#204)

Taking this into account I will close this one but for sure a fix for the compatibility and a release with it needs to be done :) 

Thank you again!

PR #202, opened 2019-10-27
PR: Add loadUiType implementation for PySide2

avalentino:
The purpose of this PR is to provide and API function (`loadUiType`) that is available in PyQt4/PyQt5 but not in PySide/PySide2.

Accordingly to what has been done for `loadUi` the new function is implemented in the `qtpy/uic.py` module and it is used only if a native implementation is not provided by the selected Qt bindings.

The `loadUiType` function load a Qt Designer .ui file and return the generated form class and the Qt base class, which is useful for customizing (e.g. adding methods) classes generated by the Qt designer.

Basic unit tests are provided as well.

PR #244, opened 2021-09-06
qtpy/tests/test_uic.py: skip if pyside2uic not installed

AndrewAmmerlaan:
Otherwise we get:

```
from qtpy import uic
/usr/lib/python3.8/site-packages/shiboken2/files.dir/shibokensupport/__feature__.py:142: in _import
return original_import(name, *args, **kwargs)
qtpy/uic.py:88: in <module>
from pyside2uic import compileUi
/usr/lib/python3.8/site-packages/shiboken2/files.dir/shibokensupport/__feature__.py:142: in _import
return original_import(name, *args, **kwargs)
E   ModuleNotFoundError: No module named 'pyside2uic'
```

This module is not guaranteed to be present since it was removed in newer versions of pyside2(-tools)

See also: https://github.com/spyder-ide/qtpy/pull/202

Signed-off-by: Andrew Ammerlaan <andrewammerlaan@gentoo.org>

### DIFF
--- a/qtpy/tests/test_uic.py
+++ b/qtpy/tests/test_uic.py
@@ -3,8 +3,12 @@ import sys
 import contextlib
 
 import pytest
-from qtpy import PYQT5, PYSIDE2, QtWidgets
+from qtpy import PYQT5, PYSIDE2, PYSIDE, QtWidgets
 from qtpy.QtWidgets import QComboBox
+
+if PYSIDE2 or PYSIDE:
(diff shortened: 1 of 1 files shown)


dalthviz (approved):
Thank you @AndrewAmmerlaan ! LGTM :+1: Will merge but I think that we need to do a proper fix to prevent `from qtpy import uic` to raise an error. I created #245 to keep track of the problem. Let us know if you want to help with that!

PR #246, opened 2021-09-10
PR: Wrap `pysideuic` and `pyside2uic` imports since they could be unavailable

### LINKED ISSUE #245: Importing `qtpy.uic` raises an exception
As mentioned in https://github.com/spyder-ide/qtpy/pull/202#discussion_r702398077 an error raises if `qtpy.uic` is imported due a missing module (`pyside2uic `). 

The error raises since there is no handling to detect if the module is available and for example do a fallback to only raise an error if `compileUi` is imported.

Thank you to @AndrewAmmerlaan for noticing this!

dalthviz:
Fixes #245

### DIFF
--- a/qtpy/uic.py
+++ b/qtpy/uic.py
@@ -81,11 +81,17 @@ else:
     if PYSIDE:
         from PySide.QtCore import QMetaObject
         from PySide.QtUiTools import QUiLoader
-        from pysideuic import compileUi
+        try:
+            from pysideuic import compileUi
+        except ImportError:
+            pass
     elif PYSIDE2:
         from PySide2.QtCore import QMetaObject
         from PySide2.QtUiTools import QUiLoader
-        from pyside2uic import compileUi
+        try:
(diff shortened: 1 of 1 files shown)


qtpy/uic.py
```diff
@@ -81,11 +81,17 @@
     if PYSIDE:
         from PySide.QtCore import QMetaObject
         from PySide.QtUiTools import QUiLoader
-        from pysideuic import compileUi
+        try:
+            from pysideuic import compileUi
+        except ModuleNotFoundError:
```
ccordoba12, line 86: `ModuleNotFoundError` is not available in Python 2. Please change it for `ImportError` instead.

ccordoba12 (commented):
@dalthviz, I left a minor comment for you to maintain compatibility with Python 2. Otherwise looks good to me.

ccordoba12 (approved):
Thanks @dalthviz!

Issue #233, opened 2021-03-17
PyQt6 support


bgermann:
Does `QtPy` support the newly released `PyQt6` package?

Related: #229

stonebig:
I see an intersting metho in PyQtGraph, but I fail to apply it in QtPy context
https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/Qt.py#L331..L345

stonebig:
PyQtGraph is pretty much leading the change https://github.com/pyqtgraph/pyqtgraph/pull/1707

The-Compiler:
FWIW the easiest approach (and what I will do in [qutebrowser](https://qutebrowser.org/) is to:

- Don't support PyQt 6.0 (I need to wait for QtWebEngine to come back with Qt 6.2 anyways), thus no need for a shim like in https://github.com/pyqtgraph/pyqtgraph/pull/1707
- Switch all enum access to the scoped variants (I wrote a quick&dirty [script to do so](https://github.com/qutebrowser/qutebrowser/issues/5904#issuecomment-736792450))
- Drop support for PyQt5 < 5.11 (mid 2018) which only allowed unscoped enum access

From what I can tell, PySide2 has always allowed scoped access. PyQt5 5.10 and earlier hasn't, and neither has PyQt4 (not sure about the Qt4 PySide). But it's really time to drop those to be honest...

j9ac9k:
pyqtgraph maintainer here; trying to maintain qt4 _and_ qt6 will be tough.  The enums namespace that PyQt6 requires I don't believe works for Qt4 bindings.  

In case it's helpful, here is where we do the bulk of our abstraction layer work (please keep in mind we don't support Qt4 any more): https://github.com/pyqtgraph/pyqtgraph/blob/master/pyqtgraph/Qt/__init__.py

tlambert03:
Yeah, I'm all for @The-Compiler's proposal too.  Including dropping support for pre 5.11

ccordoba12:
> pyqtgraph maintainer here; trying to maintain qt4 and qt6 will be tough. The enums namespace that PyQt6 requires I don't believe works for Qt4 bindings.

Hey @j9ac9k, thanks for chiming in! Ok, we'll remove support for PyQt4 before adding support for PyQt6/PySide6.

> Yeah, I'm all for @The-Compiler's proposal too. Including dropping support for pre 5.11

Ok, we'll take a look at that too.

[...]

PR #225, opened 2020-12-15
Add support for PySide6

jschueller:
Several modules are not available yet and will be re-added later in qt 6.x series:
https://doc-snapshots.qt.io/qt6-dev/whatsnew60.html#removed-modules-in-qt-6-0

PR #247, opened 2021-09-13
[for discussion] MetaPathFinder proof of principle

tlambert03:
this PR is more for the sake of discussion than anything.  It's a proof of principle for a rather dramatic re-organization of qtpy, based around a) dropping support for qt < 5.11 and python <3... and using a `MetaPathFinder` (combined with a pyi stub) instead of all of the boilerplate mock modules.  Not particularly proposing that this be used, but while discussions around pyqt6 and pyside2 are going on (along with discussions of dropping earlier versions of Qt, and perhaps dropping unscoped enum access), I figured I'd throw this out there.

The key bit, in `qtpy.__init__` is to figure out what `API_NAME` is, and then add a MetaPathFinder to sys.modules:

```python
# Setup the meta path finder that lets us import anything using `qtpy.module`
class QtPyImporter(abc.MetaPathFinder):
    def find_spec(
        self,
        fullname: str,
        path: Sequence[bytes | str] | None,
        target: ModuleType | None = None,
    ) -> ModuleSpec | None:
        """Find a spec for the specified module.

        See https://docs.python.org/3/reference/import.html#the-meta-path
        """
        if fullname.startswith(__name__):
            return util.find_spec(fullname.replace(__name__, API_NAME))
        return None

sys.meta_path.append(QtPyImporter())
```

with this you no longer need to have a bunch of  modules that look like this
```
# Local imports
from . import PYQT4, PYSIDE2, PYQT5, PYSIDE, PythonQtError

if PYQT5:
    from PyQt5.QtSvg import *
elif PYSIDE2:
    from PySide2.QtSvg import *
...
```

the remaining namespace changes can still be done in specific module overrides (i.e. if the module exists in qtpy, it will still be used), and you can provide a `pyi` file to preserve autocomplete and type stubs and stuff (this PR still needs work on the overrides and the type stubs... but it's "possible").

This strategy removes nearly all the code (2000+ lines) from the repo and all tests pass (locally, though we'll see about CI).  I have no doubt that the current PR breaks a number of things that aren't explicitly tested... but i figured i'd throw this out there for a little discussion.  

also fine to close it if it's too different to consider :)


dalthviz:
Hi @tlambert03 sorry for the late response and thank you for your interest! Regarding the changes this will indeed simplify the repo structure but I'm not totally sure if there is any shortcoming for doing such a change (like with the override procedure you mention for example or the overall understanding of the way QtPy works for new contributors). Maybe it could be nice to hear some opinions/ideas from @ccordoba12 @stonebig @j9ac9k @The-Compiler @jschueller

Just in case, the initial idea we had with @ccordoba12 for QtPy 2.0.0 was to:

* Drop Python 2 and PyQt4 support
* [Add support for PySide6](https://github.com/spyder-ide/qtpy/pull/225) and then check adding support for PyQt6 
* Check other [pending PRs](https://github.com/spyder-ide/qtpy/pulls?q=is%3Aopen+is%3Apr+milestone%3Av2.0.0)).

With the discussion at https://github.com/spyder-ide/qtpy/issues/233 the idea to just support qt > 5.11 started to seem reasonable and well here we are :)

Also, if anyone think of someone else that could have opinions/ideas let them know!

tlambert03:
Yep, makes sense!  I agree, there's no urgent "problem" that this is solving or anything.  Just a pattern that I was playing with. Definitely understand if it's not worth it :)

tlambert03:
I'll close this to keep your open PRs to a minimum :)  just wanted to document the possible approach somewhere.  cheers, Talley

PR #252, opened 2021-10-12
PR: Remove Qt4 support

jschueller:
Removes Pyside and PyQt4 support (its not even tested anymore), Qt5 is way more common nowadays and we have support for Qt6 (PySide6, and preliminary work for PyQt6)


barracuda156:
This breaks another port for me now :(

PR #259, opened 2021-10-21
PR: [PyQt6] Add exec_/print_, and add QOpenGLWidget/QShortcut to QtWidgets

### LINKED ISSUE #258: Compatibility with `QtWidgets.QOpenGLWidget` vs `QtOpenGLWidgets.QOpenGLWidget`
I think we could keep this _(the new `QtOpenGLWidgets` module in `Qt6`)_ but we need to do also the compatibility import at `QtWidgets.py` in a similar fashion as done with `PySide6`: https://github.com/spyder-ide/qtpy/blob/32f4a9f9fd9df3758c5f779fdac17132b23acc37/qtpy/QtWidgets.py#L23

What do you think @ccordoba12 ?

_Originally posted by @dalthviz in https://github.com/spyder-ide/qtpy/pull/255#r731144977_

kumattau:
PyQt6 document says
> All exec_() and print_() methods have been removed.

https://www.riverbankcomputing.com/static/Docs/PyQt6/pyqt5_differences.html

But I think it is better that QtPy supports exec_() and print_()
because the latest PySide2 5.15.2 supports exec_() only and PySide2/PySide6 support print_() only.

I think we can simply assign exec to exec_ : QApplication.exec_ = QApplication.exec (print_ is also the same)

I picked up exec_ and print_ in PySide6, and add them to corresponding classes of PyQt6.

Additionally, I added QOpenGLWidget and QShortcut to QtWidgets of PyQt6 same as PySide6.

Related to: https://github.com/spyder-ide/qtpy/issues/233#issuecomment-948885481

Fixes #258

### DIFF
--- a/qtpy/QtGui.py
+++ b/qtpy/QtGui.py
@@ -17,6 +17,7 @@ if PYQT6:
     from PyQt6.QtGui import *
     QDrag.exec_ = QDrag.exec
     QGuiApplication.exec_ = QGuiApplication.exec
+    QTextDocument.print_ = QTextDocument.print
 elif PYQT5:
     from PyQt5.QtGui import *
 elif PYSIDE2:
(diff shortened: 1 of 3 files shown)


kumattau:
I updated commit comments only

PR #268, opened 2021-10-29
PR: Fix and improve Coveralls reporting

### LINKED ISSUE #267: Coveralls shows 0 coverage despite coverage report and runner output looking fine
On PR #262 , I verified that the coverage reports and the coveralls output all looked good (after some fixes to get things working), and I can confirm they still do in the latest build. However, @dalthviz noticed that coveralls is showing zero coverage, which I suspect (given the lack of other obvious reasons) is due to it not correctly handling coverage from the installed package (vs. the local source) properly.

In addition, it would be nice if we could submit coverage from all jobs (since they cover different things), unique reports for each binding (instead of just the last one), and title the individual jobs more descriptively so its clear what platform/Python version/install type/binding they came from, though the main priority here is just getting things working again.

CAM-Gerlach:
Coveralls is now showing 0 coverage after PR #262 . This PR will fix it, as well as implement some other low-hanging fruit fixes and improvements on top of that.

As it turns out, it was indeed due to the test working dir not matching the one coveralls was running in, due to the hacky workaround to fix the issues from QtPy having a non-`src` dir layout.

Not only does this fix the issue with coveralls, but it also makes the paths much cleaner, reports coverage from all jobs and with descriptive job names, and combines coverage across all bindings (since naturally a lot of code is specific to one binding), along with fixing/cleaning up a few other things. This makes our total coverage much higher and more representative, bringing it from 25% to 75%, a 3x improvement, as well as makes the coveralls output easier to read and more informative.

* [x] Fix coveralls reporting showing zero coverage
* [x] Improve coverage reporting
    * [x] Use .coveragerc config file
    * [x] Provide a more descriptive name for the report
    * [x] Drop repo token and use Github's instead
* [x] Collect coverage from all bindings
    * [x] Use common env name for all runs and save it
    * [x] Remove common basedir from all runs
    * [x] Combine coverage from all bindings
* [x] Collect coverage from all runners

Fix #267

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,5 +103,6 @@ jobs:
         env:
           COVERALLS_REPO_TOKEN: 'xh75EzxFFMoTEyNPo3wXxXv8OVkul3eE5'
         run: |
+          cd qtpy  # Switch to test working dir per non-src-layout hack
           python3 -m pip install --upgrade coveralls
           python3 -m coveralls --service=github-actions || true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
           pyside2-version: 5.12.0  # 5.12.1-5.12.6 fails on collection/segfaults on patch test
         - os: ubuntu-latest
           python-version: '3.9'
-          use-conda: 'Yes'
+          use-conda: 'No'
           coverage: 'True'  # Collect coverage only from this run, currently
         - os: ubuntu-latest
           python-version: '3.6'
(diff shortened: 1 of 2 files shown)


CAM-Gerlach:
Sidenote: With `setuptools` 58.3.0, which just went live between when #262 was merged and now, we are now getting a formal `DeprecationWarning` when installing the package (which correctly fails the CIs, due to `-W error`) due to it using the old legacy `setup.py` setuptools backend instead of a the modern `pyproject.toml`-based setuptools one conforming to the PEP 517 standard. I was already planning to migrate in a followup PR to #262 since its pretty straightforward to do and I'm well familiar with it, but this gives it more urgency. For now I'll just drop `-W error` so we can get this PR merged and handle it in the next one.

CAM-Gerlach:
@dalthviz As usual, I spent way more time on this than I should have, but this brings our reported coverage from the pre- #262 25% to now 75%, a full 3x improvement, with more detailed, representative and easy to read data. We might want to consider adding coveralls to our formal CI checks, like Spyder has, to check for significant regressions.

dalthviz (approved):

CAM-Gerlach:
Thanks @dalthviz

PR #272, opened 2021-11-01
PR: Upgrade packaging to avoid legacy/deprecated behavior and follow PEP 517

### LINKED ISSUE #270: Port packaging from deprecated legacy builder to current PEP 517 standards
I was already planning to do this following #262 being merged, as I've done for many other packages. However, with `setuptools` 58.3.0, which we are now getting a formal `DeprecationWarning` when installing the package (which correctly fails the CIs, due to `-W error`), and it will soon stop working entirely.

The issue is due to using the old legacy `setup.py` backend that has been discouraged for a number of years, to the modern `pyproject.toml`-based setuptools backend that conforms to the PEP 517 standard. Its straightforward to migrate and should no negative/backward compatibly effect on users (other than improving the robustness of building and installing the package), as PEP 517-conformant builds actually allows installing even on machines that don't have an otherwise compatible installation of setuptools with suitably recent pip, and I'll still include a backward-compatibly setuptools shim for older ones, so it'll still work as before for older non-PEP 517  pip/wheel/setuptools. And of course, this is all now fully tested end to end on all platforms and versions in our CIs, following #262 being merged.

So, I'll migrate the config from the dynamic `setup.py` to the determistic `setup.cfg` and simplify it in the process (e.g. no need for bespoke code to get the version, readme, etc, as that is all handled by setuptools at build time), add a setup.py shim for backward compatibility, and add a pyproject.toml with the basic PEP 517 config, all following the standard PyPA recommendations as now being followed by the rest of the ecosystem. I'll also update the CIs and the docs where applicable.

_Originally posted by @CAM-Gerlach in https://github.com/spyder-ide/qtpy/issues/268#issuecomment-955067333_

CAM-Gerlach:
Currently, the package is using the deprecated legacy builder and calling setup.py directly, which is causing warnings and will eventually go away completely. This migrates it to the (more) modern PEP 517 setuptools builder and setup.cfg, with a setup setup.py for backward compatibility. This also makes one change to how the version is handled to avoid having to import the package or execute any dynamic code to get the version statically, which shouldn't affect anything else as `_version` is marked as internal use only (with leading `_`). I've compared the wheels and sdists and, aside from the minor updates to the metadata made in a separate commit, they are byte for byte identical.

* [x] Update packaging for initial compatibility with PEP 517 and PEP 518
    * [x] Add pyproject.toml
    * [x] Update CIs
    * [x] Update RELEASE.md accordingly
* [x] Use modern declarative setup.cfg instead of legacy dynamically-executed setup.py
    * [x] Migrate setup.py metadata to setup.cfg
    * [x] Fix/avoid encoding issues on open()
    * [x] Run setup-cfg-format
    * [x] Add stub setup.py for backward compat with legacy tools
    * [x] More reliable static __version__ handling
    * [x] Update RELEASE,md accordingly
* [x] Add a few new bits of setup.cfg metadata
    * [x] Add additional appropriate Trove tags
    * [x] Add project urls
    * [x] Add test extra
* [x] Update RELEASE,md to follow modern PEP 517/518 best practices

Resolve #270

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,31 +1,58 @@
-To release a new version of qtpy on PyPI:
+# Release Procedure
 
-* Close Github milestone
+To release a new version of QtPy on PyPI:
 
-* git fetch upstream && git merge upstream/master
 
-* git clean -xfdi
+## Prepare
 
-* Update CHANGELOG.md using `loghub` to generate the list of issues and PRs merged to add at the top of the file
+* Close Github milestone and ensure all issues are resolved/moved
 
-    loghub -m vX.X.X spyder-ide/qtpy
+* `git switch master && git pull upstream master` (assuming `master` is the release branch)
 
-* Update `__version__` in `__init__.py` (set release version, remove 'dev0')
+* Check `MANIFEST.in` and `setup.cfg` to ensure they are up to date
 
-* git add . && git commit -m 'Release X.X.X'
 
-* python -m build
+## Commit
 
-* twine check dist/*
+* `pip install --upgrade loghub`
 
-* twine upload dist/*
+* Update `CHANGELOG.md` using `loghub` to generate the list of issues and PRs merged to add at the top of the file:
 
-* git tag -a vX.X.X -m 'Release X.X.X'
+  ```bash
+  loghub -m vX.Y.Z spyder-ide/qtpy
+  ```
 
-* Update `__version__` in `__init__.py` (add 'dev0' and increment minor)
+* Update `__version__` in `__init__.py` (set release version, remove `.dev0`)
 
-* git add . && git commit -m 'Back to work'
+* `git commit -am 'Release X.Y.Z'`
 
-* git push
 
-* git push --tags
+## Build
+
+**Note**: We use `pip` instead of `conda` here even on Conda installs, to ensure we always get the latest upstream versions of the build dependencies.
+
+* `git clean -xfdi`
+
+* `python -m pip install --upgrade-strategy eager --upgrade build pip setuptools twine wheel`
+
+* `python -bb -X dev -W error -m build`
+
+
+## Upload
+
+* `twine check --strict dist/*`
+
+* `twine upload dist/*`
+
+* `git tag -a vX.Y.Z -m 'Release X.Y.Z'`
+
+
+## Cleanup
+
+* Update `__version__` in `__init__.py` (add `.dev0` and increment minor)
+
+* `git commit -am 'Back to work'`
+
+* `git push upstream --follow-tags`
+
+* Create a GitHub release from the tag
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -14,9 +14,7 @@ To release a new version of qtpy on PyPI:
 
 * git add . && git commit -m 'Release X.X.X'
 
-* python setup.py sdist
-
-* python setup.py bdist_wheel
+* python -m build
 
 * twine check dist/*
 
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -10,7 +10,7 @@ To release a new version of qtpy on PyPI:
 
     loghub -m vX.X.X spyder-ide/qtpy
 
-* Update `_version.py` (set release version, remove 'dev0')
+* Update `__version__` in `__init__.py` (set release version, remove 'dev0')
 
 * git add . && git commit -m 'Release X.X.X'
 
@@ -22,7 +22,7 @@ To release a new version of qtpy on PyPI:
 
 * git tag -a vX.X.X -m 'Release X.X.X'
 
-* Update `_version.py` (add 'dev0' and increment minor)
+* Update `__version__` in `__init__.py` (add 'dev0' and increment minor)
 
 * git add . && git commit -m 'Back to work'
 
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -61,7 +61,7 @@ import sys
 import warnings
 
 # Version of QtPy
-from ._version import __version__
+__version__ = '2.0.0.dev0'
 
 
 class PythonQtError(RuntimeError):
--- a/setup.cfg
+++ b/setup.cfg
@@ -16,23 +16,33 @@ license_files =
     LICENSE.txt
 classifiers =
     Development Status :: 5 - Production/Stable
+    Environment :: MacOS X
     Environment :: Win32 (MS Windows)
     Environment :: X11 Applications :: Qt
     Intended Audience :: Developers
     License :: OSI Approved :: MIT License
     Operating System :: OS Independent
     Programming Language :: Python :: 3
+    Programming Language :: Python :: 3 :: Only
     Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
+    Topic :: Software Development :: Libraries
+    Topic :: Software Development :: User Interfaces
+    Topic :: Software Development :: Widget Sets
 keywords = qt PyQt5 PyQt6 PySide2 PySide6
+project_urls =
+    Github = https://github.com/spyder-ide/qtpy
+    Bug Tracker = https://github.com/spyder-ide/qtpy/issues
+    Parent Project = https://www.spyder-ide.org/
 
 [options]
 packages = find:
 install_requires =
     packaging
 python_requires = >=3.6
+include_package_data = True
 zip_safe = False
 
 [options.packages.find]
@@ -40,3 +50,8 @@ exclude =
     contrib
     docs
     tests*
+
+[options.extras_require]
+test =
+    pytest>=6.0.0,<7.0
+    pytest-cov>=2.11.0
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,0 +1,42 @@
+[metadata]
+name = QtPy
+version = attr: qtpy.__version__
+description = Provides an abstraction layer on top of the various Qt bindings (PyQt5/6 and PySide2/6) and additional custom QWidgets.
+long_description = file: README.md
+long_description_content_type = text/markdown
+url = https://github.com/spyder-ide/qtpy
+author = Colin Duquesnoy and the Spyder Development Team
+author_email = spyder.python@gmail.com
+maintainer = Spyder Development Team and QtPy Contributors
+maintainer_email = spyder.python@gmail.com
+license = MIT
+license_file = LICENSE.txt
+license_files =
+    AUTHORS.md
+    LICENSE.txt
+classifiers =
+    Development Status :: 5 - Production/Stable
+    Environment :: Win32 (MS Windows)
+    Environment :: X11 Applications :: Qt
+    Intended Audience :: Developers
+    License :: OSI Approved :: MIT License
+    Operating System :: OS Independent
+    Programming Language :: Python :: 3
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+keywords = qt PyQt5 PyQt6 PySide2 PySide6
+
+[options]
+packages = find:
+install_requires =
+    packaging
+python_requires = >=3.6
+zip_safe = False
+
+[options.packages.find]
+exclude =
+    contrib
+    docs
+    tests*
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -24,6 +24,7 @@ jobs:
       CI: 'True'
       PYTHON_VERSION: ${{ matrix.python-version }}
       USE_CONDA: ${{ matrix.use-conda }}
+      SKIP_PIP_CHECK: ${{ matrix.skip-pip-check }}
       PYQT5_VERSION: ${{ matrix.pyqt5-version }}
       PYQT6_VERSION: ${{ matrix.pyqt6-version }}
       PYSIDE2_VERSION: ${{ matrix.pyside2-version }}
@@ -43,6 +44,7 @@ jobs:
           special-invocation: 'xvfb-run --auto-servernum '
         - python-version: '3.6'
           use-conda: 'No'
+          skip-pip-check: 'true'  # Pytest packaging issue on Python 3.6 defaults channel
           pyside2-version: 5.12.0  # 5.12.1-5.12.6 fails on collection/segfaults on patch test
         - os: ubuntu-latest
           python-version: '3.6'
@@ -104,6 +106,5 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           cd qtpy  # Switch to test working dir per non-src-layout hack
-          python -m pip install --upgrade coveralls
           cat qtpy_basedir.txt
-          python -b -X dev -m coveralls --service=github --rcfile="../.coveragerc" --basedir="$(cat qtpy_basedir.txt)"
+          pipx run coveralls --service=github --rcfile="../.coveragerc" --basedir="$(cat qtpy_basedir.txt)"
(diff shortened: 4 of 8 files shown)


setup.cfg
```diff
@@ -0,0 +1,58 @@
...
+    Github = https://github.com/spyder-ide/qtpy
+    Bug Tracker = https://github.com/spyder-ide/qtpy/issues
+    Parent Project = https://www.spyder-ide.org/
+
+[options]
+packages = find:
+install_requires =
+    packaging
+python_requires = >=3.6
+include_pacakge_data = True
```
Kant-ain, line 45:
```suggestion
include_package_data = True
```

setup.cfg
```diff
@@ -0,0 +1,58 @@
...
+keywords = qt PyQt5 PyQt6 PySide2 PySide6
+project_urls =
+    Github = https://github.com/spyder-ide/qtpy
+    Bug Tracker = https://github.com/spyder-ide/qtpy/issues
+    Parent Project = https://www.spyder-ide.org/
+
+[options]
+packages = find:
+install_requires =
+    packaging
```
dalthviz, line 43: Should this have the constraint `>=20.0` and maybe be removed for the `extras_require::test`?

dalthviz (commented):
Thanks @CAM-Gerlach ! I left a suggestion and a comment/question. Besides that this LGTM :+1:

Also, thank you @Kant-ain for the misspelling fix suggestion !

qtpy/__init__.py
```diff
@@ -61,7 +61,7 @@
 import warnings
 
 # Version of QtPy
-from ._version import __version__
+__version__ = '2.0.0.dev0'
```
ccordoba12, line 64: I don't agree with this change because the previous way to set `__version__` is what we use in all our other projects. Its merits or flaws can be discussed in another PR if you want, but for now please leave things as they were, i.e.

```python
version_info = (2, 0, 0, 'dev0')
__version__ = '.'.join(map(str, version_info))
```

RELEASE.md
```diff
@@ -1,33 +1,65 @@
...
-* git fetch upstream && git merge upstream/master
 
-* git clean -xfdi
+## Prepare
 
-* Update CHANGELOG.md using `loghub` to generate the list of issues and PRs merged to add at the top of the file
+* Close Github milestone and ensure all issues are resolved/moved
 
-    loghub -m vX.X.X spyder-ide/qtpy
+* `git switch <release-branch>` then `hub sync` or `git fetch upstream && git merge upstream/<release-branch>`
```
ccordoba12, line 10:
```suggestion
* `git fetch upstream && git merge upstream/master`
```

* Instructions should be as simple as possible (just copy/paste), so no need to add two methods here and specially one that requires another tool besides git.
* `1.x` is an exceptional case of a release branch. Most probably in the future we won't maintain others.

ccordoba12 (commented):
Thanks @CAM-Gerlach for this! I left a couple of comments, otherwise looks good to me.

setup.cfg
CAM-Gerlach: Thanks for the catch! I actually noticed it at the time, but I must have fixed it in the wrong file or not saved it. Rebased the change in, since I moved this from the second to the third commit for clarity, so its fixed now.

RELEASE.md
```diff
@@ -1,33 +1,65 @@
...
+* `python -bb -X dev -W error -m pip install -e .`
+
+
+## Upload
+
+* `twine check --strict dist/*`
+
+* `twine upload dist/*`
+
+* `git tag -a vX.Y.Z -m 'Release X.Y.Z'`
```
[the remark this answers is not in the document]
CAM-Gerlach: I modified this because `X.X.X` implies that the the version components are the same variable repeated (e.g. 1.1.1), whereas `X.Y.Z` makes it clear that they are three different variables, and that the two different uses of the X.Y.Z string _are_ equivalent. Furthermore, this is the most widely used standard convention (at least in English) for describing arbitrary version and tag names, e.g. in [SemVer](https://semver.org/) itself, and a search for `version "X.Y.Z"` gives 400 million hits for mostly version-related things, while `version "X.X.X"` gives you 600 million hits that are overwhelmingly dominated by...well, something that isn't very family-friendly (one more reason to avoid it, heh).

setup.cfg
CAM-Gerlach: Good question; in this case the constraint is a soft one, and only has potential effect on its use in development/packaging/testing to be fully certain of consistent results, rather than something that matters for the very narrowly-scoped usage at runtime. But its not critical either way, it can be removed under the `extra` if you want.

RELEASE.md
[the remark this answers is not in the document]
dalthviz: Then we should use the same convention in the other commands above (they still are using the `X.X.X` convention)

setup.cfg
dalthviz: Then please go ahead and remove it from the `extra` section

RELEASE.md
CAM-Gerlach: Sure, fair enough on both. I wasn't sure about including both to begin with, and it actually makes it sufficiently ambigous that even you were unclear that `git switch master` was supposed to apply to both. Before I go ahead with this, should we keep the other couple lines that offers two methods (`pip` and `conda`) as they are, or should we remove the `pip` alternative and only keep `conda` there too?

CAM-Gerlach: Also in the spirit of simplification, is there a reason you specify `git fetch upstream && git merge upstream/master`, when `git pull upstream master` is equivalent? Per the [docs](http://git-scm.com/docs/git-pull),

> In its default mode, `git pull` is shorthand for `git fetch` followed by `git merge FETCH_HEAD`.

(Since `FETCH_HEAD` in the case of the given `pull` command is `upstream/master`)

qtpy/__init__.py
CAM-Gerlach: I made every attempt to keep things exactly as they were here, but as `__version__` is not a static attribute and requires dynamic runtime code execution, `setuptools` fails to statically retrieve it from the AST during installation without importing the package and executing arbitrary code, which in turns triggers the complex, environment-dependent logic in `__init__.py` (which we _definitely_ don't want running on install) and crashes due to missing third party imports which haven't been installed yet. Furthermore, I tested defining a static `__version__` in `_version.py` and keeping everything else as it is, but it still triggers executing the `__init__` as before, and thus the same crash on install.

So, the simplest solution for this PR that didn't require risky hacks or much bigger changes (e.g. using `setuptools_scm`, a VERSION file, or another third-party tool) was to simply include `__version__` in `__init__.py` directly, particularly given the motivation of a separate `_version.py` to begin with was to allow the (now-legacy) setup.py to load and execute it, which is nowadays unnecessary and deprecated, and is exactly the behavior this PR migrates away from. See pypa/setuptools#1724 for more details (there were a few other workarounds there, but they require `__init__.py` to not have complex logic or third-party package imports, or rely on implementation-dependent hacks, and in any case didn't work in my testing).

RELEASE.md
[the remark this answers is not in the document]
CAM-Gerlach: Ah, great point—don't know how I missed those. I went back and did so, thanks.

setup.cfg
CAM-Gerlach: Sure, done.

RELEASE.md
CAM-Gerlach: I went ahead and simplified this line to just `git switch master && git pull upstream master`; lmk what you want me to do about the `pip` and `conda` alternatives

qtpy/__init__.py
ccordoba12: Ok, I see, but how does this work?

> a VERSION file

I mean, that sounds similar to what we had, doesn't it?

RELEASE.md
ccordoba12: > lmk what you want me to do about the pip and conda alternatives

What do you mean by this?

ccordoba12: Oh, I see. Yes, we prefer to use `pip` to do releases because the build and upload dependencies are always up to date.

So please remove the commands related to conda and leave a mention saying why we prefer to use pip.

qtpy/__init__.py
CAM-Gerlach: Yeah, on the surface it may seem that way, but on closer investigation its actually quite different.

What we had before was an internal only (`_version.py`) Python module that both `__init__.py` could `import` and `setup.py` could `exec`.

The VERSION file I was referring to is a _plain text file_ (as typically used by third-party version management tools) with the version as a string, which `setup.cfg` would read in with `file:`, but for `__init__.py` to read it, we'd need to correctly include it in `package_data` and/or `MANIFEST.in` (with `include_package_data=True`) to ensure it was included in both sdists and wheels (or the package would crash on import), _and_ then open, read and parse it in `__init__,py` using `importlib.resources`/`importlib_resources`, which itself would either require unnecessarily dropping Python 3.6 support, or adding another third party dependency, increasing further the complexity and cost.

Given that we can now just include the `__version__` string right in `__init__.py` and `setup.cfg` based setuptools can simply read it directly from there with `attr:qtpy.__version__` without executing the file, without adding any of the complex, potentially risky gymnastics that the `file:` approach would require, this clearly seemed to be the simplest and most comparable working alternative.

RELEASE.md
CAM-Gerlach: Sorry, I missed seeing this reply somehow.

Makes sense; I removed the conda command and the separate install step (since `pip install --upgrade` does both).

I also changed the `pip` invocation to use `python -m pip` to be cross-platform compatible, moved the `git clean` step to right before the build, where it belongs (and to balance out the sections a bit), and added an explict `pip install --upgrade loghub` command to install/upgrade loghub.

CAM-Gerlach: Also, made the Pytest invocation in the release checklist more consistent and robust, and (in a separate commit) added `pip check` to the release checklist, and `pip check` and `twine check` to the CIs.

qtpy/__init__.py
CAM-Gerlach: Hey @ccordoba12 this is the last open review comment on this PR; is this okay to merge now?

hiaselhans: i can't seem to find any use of `version_info` in the lib, but if it's of any importance, why not just doing it the other way around:

``` python
__version__ = "2.0.0.dev0"
version_info = __version__.split(".")
```

CAM-Gerlach: Yep, thanks; I actually thought of that too in case any of our packages actually do use this value. In this case, it was never imported/exposed in `__init__` (or anywhere else public) to begin with, just in its own private internal-only module, so I didn't do that here. If for some reason we do want to, though, since its a constant in the top-level package namespace and wasn't here before, it would make sense to make it `VERSION_INFO`.

The issue with this (and equally, the old `version_info` -> `__version__` approach) versus just using `__version__` (which is what virtually all third party packages are likely to look for anyway, and has no real use internally) is that in the canonical PEP 440-normalized form, there isn't supposed to be a `.` between the last version component and alpha/beta/rc tags, only for  `devN`. This will get normalized in the canonical version in the package metadata, but is strongly discouraged and means the version stored in the package metadata, used by packaging tools and exposed via `importlib.metadata`, `pkg_resources`, etc will be inconsistent with the version in `__version__`. While this isn't a regression over the old approach, using `__version__` would free us from that limitation.

ccordoba12: > this clearly seemed to be the simplest and most comparable working alternative.

Hey @CAM-Gerlach, sorry for my late reply. Although I don't like too much, I think leaving things as they are right now is the best option.

RELEASE.md
```diff
@@ -1,33 +1,69 @@
...
+* `git clean -xfdi`
+
+* `python -m pip install --upgrade-strategy eager --upgrade build pip setuptools twine wheel`
+
+* `python -bb -X dev -W error -m build`
+
+
+## Test
+
+* `python -bb -X dev -W error -m pip install dist/*.whl`
```
ccordoba12, line 43: Does this install the newly created wheel in your environment? If that's the case, why do we want that?

RELEASE.md
```diff
@@ -1,33 +1,69 @@
...
+
+* `pip check`
+
+* `python -bb -X dev -W error -m pytest qtpy/tests`
```
ccordoba12, line 47: Why this? We don't run our tests locally after creating the wheel and tarball were created.

RELEASE.md
```diff
@@ -1,33 +1,69 @@
...
+
+* `python -bb -X dev -W error -m pip install -e .`
```
ccordoba12, line 49: What's the purpose of this?

RELEASE.md
CAM-Gerlach: Yup. Otherwise, the final release version is never actually tested anywhere before going live to the world, either to see if it can be installed successfully, that the installed environment is consistent (`pip check`) or that tests pass after installing. Even if there are no code changes, there could be a packaging error/typo, something in the packaging stack could have been updated since the last PR (which in fact happened on PR #268 and errored the install), or a user mistake somewhere in the release process. Even if the chances are low, the consequences are very high (especially given how many other packages/applications depend on QtPy) so its worth spending just a few extra seconds per release to do a sanity check that it actually works.

The alternative is creating a branch and submitting the code after line 27 as a PR, which would test it on all platforms in a clean env (and is what I do), but that wouldn't match your current standard workflow and take considerably more time (minutes rather than seconds) for diminishing returns, so I didn't suggest it here.

RELEASE.md
CAM-Gerlach: Well, if not, then it risks releasing a broken package to the world, which considering QtPy is used by at least 56 300 other projects on GitHub (not counting Gitlab, Bitbucket, SourceForge, other platforms, local development, implicit deps and those dependabot can't read, which is a lot of them) and is top 200 on PyPI, is not a trivial thing :)

RELEASE.md
CAM-Gerlach: Reinstall the dev version locally after installing and testing the wheel. (Also, incidentally, verifies it installs successfully in dev mode)

qtpy/__init__.py
CAM-Gerlach: No worries, I know you have a lot going on and sorry for bugging you.

Okay, could you clarify what you mean by "leave things as they are"; do you mean leave things as they are in the PR or as they are in the repo (i.e. not merging the PR, or at least reverting the core improvements, and doing it over again later at some point)? Thanks!

ccordoba12: I mean leaving `__version__` in `__init__.py` as it's currently implemented in this PR.

RELEASE.md
ccordoba12: > The alternative is creating a branch and submitting the code after line 27 as a PR

I think this is really unnecessary because after the last PR merged in the repo and before the release, there are zero code changes (except for changing the package version). So we can be certain that our tests are passing.

In all my years as a maintainer, I have released no more than 10 broken packages in hundreds of releases. Also, since this adds more work to maintainers, I'm -1 on it (you'd have to recreate the testing env locally and keep it updated, when it's running perfectly fine on Github).

> Even if the chances are low

I think they are.

> the consequences are very high

No. You can always yank the release on PyPI.

> so its worth spending just a few extra seconds per release to do a sanity check that it actually works.

Seconds if we release frequently. But it could take much more if releases are infrequent.

ccordoba12: Since this is a change in our release process and it has nothing to do with this PR (whose purpose is to use the new way to create packages), I suggest to remove this section and discuss it in a separate issue. That way we could quickly move forward with this.

CAM-Gerlach: > Since this is a change in our release process and it has nothing to do with this PR (its purpose is to use the new way to create packages), I suggest to remove this section and discuss it in a separate issue. That way we could quickly move forward with this.

Sure, more than fair. I updated the PR accordingly to remove the ``Test`` section from the Release guide, so this should be ready.

dalthviz:
Thanks @CAM-Gerlach !!

CAM-Gerlach:
Thanks for your feedback and bearing with me on this @ccordoba12 and @dalthviz !

PR #273, opened 2021-11-03
PR: Prevent future deprecations messages and decrease general technical debt + code style issues

CAM-Gerlach:
Fixes a number of minor problems in the codebase, including:

* [x] Replace the deprecated `tmpdir` pytest fixture which uses legacy py.path objects with `tmp_path`, which uses standard `pathlib.Path`s (and rename helper function arg to avoid confusion)
* [x] Use explicit `encoding` on `open()`, to avoid platform/environment-dependent bugs, deprecation warning on Py3.10 and eventual removal
* [x] Fix other correctness issues, including:
    * [x] Bad indent levels in some lines/blocks
    * [x] Missing `raise from` on re-raised exceptions to re-raise correctly
    * [x] Bare except that will trap KeyboardInterrupt, termination and other unwanted errors
    * [x] Misplaced strings intended as docstrings but not the first line in their module/callable

### DIFF
--- a/qtpy/QtCharts.py
+++ b/qtpy/QtCharts.py
@@ -12,15 +12,19 @@ from . import PYQT5, PYQT6, PYSIDE2, PYSIDE6, PythonQtError
 if PYQT5:
     try:
         from PyQt5.QtChart import *
-    except ImportError:
-        raise PythonQtError('The QtChart module was not found. '
-                            'It needs to be installed separately for PyQt5.')
+    except ImportError as error:
+        raise PythonQtError(
+            'The QtChart module was not found. '
+            'It needs to be installed separately for PyQt5.'
+            ) from error
 elif PYQT6:
     try:
         from PyQt6.QtCharts import *
-    except ImportError:
-        raise PythonQtError('The QtCharts module was not found. '
-                            'It needs to be installed separately for PyQt6.')
+    except ImportError as error:
+        raise PythonQtError(
+            'The QtCharts module was not found. '
+            'It needs to be installed separately for PyQt6.'
+            ) from error
 elif PYSIDE6:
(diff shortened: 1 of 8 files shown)


dalthviz:
Hi @CAM-Gerlach thanks for checking into this! However, just to understand better some of the changes here, the deprecations (`pytest` fixture, open with `enconding` kwarg) are related with Python 3.10?

CAM-Gerlach:
@dalthviz Sure:

* The Pytest `tmpdir` fixture has been targeted for removal in Pytest for some time as it relies on the long-deprecated `py`  package and legacy `py.path` objects; there's no connection to Python version. See pytest-dev/pytest#7259
* The lack of an explicit `encoding` kwarg on `open` will become an optional warning in Python 3.10, with the plan to be a `DeprecationWarning` in future Python versions, and possibly eventually an error. However, the problem it fixes is present in all past and present Python versions, namely that the current default encoding is platform, environment and user-dependent  (e.g. on Windows, some Linux and other platforms it is _not_ UTF-8, as many developers assume, and can be anything, potentially not even ASCII-compatible) and is unsafe to rely upon, as attempting to open files in an arbitrary encoding is liable to cause an error or undesired behavior. See [PEP 597](https://www.python.org/dev/peps/pep-0597/) for more details (N.B., I helped some with the PEP and docs, and a bit with the implementation).
* The other changes are all relatively smaller but non-trivial correctness issues that have actual runtime impact and don't directly involve a deprecation, they've been problems _ab initio_

dalthviz (approved):

CAM-Gerlach:
Thanks @dalthviz ; revised the PR title again to fix some introduced grammar and clarity issues

PR #276, opened 2021-11-08
PR: Fix regression in FORCE_QT_API behavior from merging PySide6 support

tlambert03:
#225 somehow reverted the fix for the broken `FORCE_QT_API` that was fixed in #230.  presumably a merge fail... not sure if there are others

on current master (this pretty much destroys environments with multiple Qt backends)

```py
In [1]: import PySide6

In [2]: from qtpy import QtCore

In [3]: QtCore.QTimer
Out[3]: PyQt6.QtCore.QTimer
```

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -100,9 +100,8 @@ is_old_pyqt = is_pyqt46 = False
 PYQT5 = True
 PYQT6 = PYSIDE2 = PYSIDE6 = False
 
-# When `FORCE_QT_API` is set, we disregard
-# any previously imported python bindings.
-if 'FORCE_QT_API' in os.environ:
+# Unless `FORCE_QT_API` is set, we use previousl imported python bindings.
+if not os.environ.get('FORCE_QT_API'):
     if 'PyQt6' in sys.modules:
         API = initial_api if initial_api in PYQT6_API else 'pyqt6'
     elif 'PyQt5' in sys.modules:
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -100,7 +100,7 @@ is_old_pyqt = is_pyqt46 = False
 PYQT5 = True
 PYQT6 = PYSIDE2 = PYSIDE6 = False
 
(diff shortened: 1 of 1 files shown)


qtpy/__init__.py
```diff
@@ -100,9 +100,8 @@ class PythonQtWarning(Warning):
 PYQT5 = True
 PYQT6 = PYSIDE2 = PYSIDE6 = False
 
-# When `FORCE_QT_API` is set, we disregard
-# any previously imported python bindings.
-if 'FORCE_QT_API' in os.environ:
+# Unless `FORCE_QT_API` is set, we use previousl imported python bindings.
```
tlambert03, line 103:
```suggestion
# Unless `FORCE_QT_API` is set, we use previously imported python bindings.
```

CAM-Gerlach (approved):
Good catch, @tlambert03 , thanks! LGTM, aside from the second commit needing a `fixup` (and only if @dalthviz would rather not squash-merge it).

I was going to comment on the introduced typo (and clarify the comment a bit), but you fixed the former faster than I could do so, heh.

qtpy/__init__.py
```diff
@@ -100,9 +100,8 @@ class PythonQtWarning(Warning):
 PYQT5 = True
 PYQT6 = PYSIDE2 = PYSIDE6 = False
 
-# When `FORCE_QT_API` is set, we disregard
-# any previously imported python bindings.
-if 'FORCE_QT_API' in os.environ:
+# Unless `FORCE_QT_API` is set, we use previously imported python bindings.
```
tlambert03, line 103:
```suggestion
# Unless `FORCE_QT_API` is set, use previously imported Qt Python bindings
```

CAM-Gerlach:
NB @dalthviz , the second and third commit either needs a `fixup` into the first, or the PR can simply be squash-merged, to clean up the duplicate, trivial commits.

dalthviz (approved):
Thanks @tlambert03 for noticing this and submitting a fix! LGTM :+1:

Issue #85, opened 2016-10-27
Add basic project documentation and example usage


goanpeca:
Written and edited by @CAM-Gerlach (OP was initially empty):

* [ ] Add/move general information currently contained in `__init__` docstring to the README for visibility
* [ ] Clarify that QtPy follows PyQt5 conventions, and link Qt and PyQt5 API docs
* [ ] Add examples of using QtPy in place of the specific bindings
* [ ] List/describe the methods/helpers specific to QtPy (listed in #61 )
    * [ ] `from qtpy.QtCore import Signal, Slot`
    * [ ] `QFileDialog` helpers
    * [ ] `qtpy.QtCore.__version__`
    * [ ] Others?
* [ ] Document the QtPy-specific environment variables (`FORCE_QT_API`, etc)
* [ ] Describe the main package-level constants (`PYQT5`, `PYSIDE2`, etc)
* [ ] Clarify the benefits of using QtPy with regard to cross-Qt-version migration, and link success stories (described in #62  )
    * [ ] git-cola/git-cola#232
    * [ ] [Glue](http://docs.glueviz.org/en/stable/developer_guide/qt_development.html#using-qtpy)
* [ ] Clarify Qt version compatibility in the readme

SylvainCorlay:
FYI, readthedocs is not very good with dependencies like qt. But it now supports a conda `environment.yml`. Check out qtawesome's rtd configuration (`readthedocs.yml` + `environment.yml`).

AndydeCleyre:
Do the docs exist in some format?

machinekoder:
A simple example in the README would work as well.

dalthviz:
Related issues:

* https://github.com/spyder-ide/qtpy/issues/61
* https://github.com/spyder-ide/qtpy/issues/62
* https://github.com/spyder-ide/qtpy/issues/66

dalthviz:
@CAM-Gerlach would you like to work on this?

CAM-Gerlach:
I can't promise anything like a full-on docs site like we have for `spyder-docs` or the API docs (at least in the near term), but I was actually already planning to at least update and expand the README and add a contributing/dev guide to address some of these issues (and replied in each of them with more detail).

[...]

PR #279, opened 2021-11-08
PR: Add back legacy PYQT4 and PYSIDE package-level constants for compat

CAM-Gerlach:
As discussed in [#85 (comment)](https://github.com/spyder-ide/qtpy/issues/85#issuecomment-963393873), especially if we're going to keep around old non-constant top-level variables that were only used internally (`is_old_pyqt` and `is_pyqt46`), we may as well keep the `PYQT4` and `PYSIDE` module-level constants around (always `False` now) for backward compatibility, and easier cross-compat between QtPy versions.


dalthviz (approved):
Thanks @CAM-Gerlach ! LGTM 👍🏼

PR #281, opened 2021-11-17
PR: Catch `PythonQtError` when trying to do alias for `QtDataVisualization` and dev version correction

### LINKED ISSUE #280: import qtpy raise except when only have PyQt4
https://github.com/spyder-ide/qtpy/blob/61e84329ecd5bfb3a6ff822fa6ce9758be23516c/qtpy/__init__.py#L197

```
>>> import qtpy
  File "/python3.8/site-packages/qtpy/__init__.py", line 237, in <module>
    from . import QtDataVisualization as QtDatavisualization
  File "/python3.8/site-packages/qtpy/QtDataVisualization.py", line 22, in <module>
    raise PythonQtError('No Qt bindings could be found')
```

Why you remove the QT_API check before import QtDataVisualization since 1.11.2?

dalthviz:
Fixes #280

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -235,5 +235,5 @@ try:
     # QtDataVisualization backward compatibility (QtDataVisualization vs. QtDatavisualization)
     # Only available for Qt5 bindings > 5.9 on Windows
     from . import QtDataVisualization as QtDatavisualization
-except ImportError:
+except (ImportError, PythonQtError):
     pass
--- a/qtpy/_version.py
+++ b/qtpy/_version.py
@@ -1,2 +1,2 @@
-version_info = (2, 0, 0, 'dev0')
+version_info = (1, 12, 0, 'dev0')
 __version__ = '.'.join(map(str, version_info))


ccordoba12 (approved):

CAM-Gerlach:
FYI @dalthviz , this was milestoned for v1.11.3, but the version attribute was changed to `1.12.0`...which one did you intend?

dalthviz:
Hi @CAM-Gerlach `1.12.0.dev0` is the dev version for the branch, on release this will change to be `1.11.3` and then add the end of the release will go back to `1.12.0.dev0` and no further release will be done from that branch (it is a similar approach as the one in Spyder with the `4.x` branch where the dev version has the incremented minor i.e `4.3.0.dev0` and the latest release was `4.2.5`)

CAM-Gerlach:
Huh, that wasn't how I remember it used to be with Spyder before. Unless I'm misunderstanding something, its kinda strange to use a dev version that is knowingly higher than the intended release version from the same branch _before_ that version is released, such that the version of the previous dev version will sort _higher_ than the future release version from the same branch, i.e. a current checkout would be considered a _later_ version despite coming _before_ the 1.11.3 release to which it is actually an _ancestor_. Would seem to violate valid PEP 440 versioning practices, though the impact is limited by the fact that its only a dev version.

PR #282, opened 2021-11-17
PR: Update README sponsors section

dalthviz:


CAM-Gerlach:
Hey @dalthviz , was this intended to be a draft PR? Looks like the inline image isn't rendering; otherwise, everything LGTM so far :+1:

ccordoba12 (commented):
Please add the Numfocus logo next to the Quansight one (as we do in our other repos).

Also, is there a Quasight logo with a transparent background we could use? I think it'd be nicer than having one with white background.

CAM-Gerlach (approved):
LGTM, thanks @dalthviz !

ccordoba12 (approved):
Looks good to me now, thanks @dalthviz!

PR #283, opened 2021-11-17
PR: Add a warning for developers still running legacy Qt4-based APIs

### LINKED ISSUE #261: Add a deprecation warning for unsupported `Qt` versions and bindings (at least `Qt4`: `PyQt4` and `PySide`)
Maybe could be worthy to release `1.11.3` with a depreciation warning:

- Deprecation warning for `Qt4` (removed `PyQt4` and `PySide` support for `v2.0.0`)
- Deprecation warning for older versions of `Qt5`: Depending on what minimum `Qt5` and bindings versions we will support for `v2.0.0`. Maybe `5.9` (if we can make it work), but most probably `5.12`

CAM-Gerlach:
We will now warn developers (via a `DeprecationWarning`, so they have to be running Python with `-X dev`, `-W default` or `-W error`, or import QtPy in their `__main__` script) when they are using long-deprecated, soon to be removed Qt4 based APIs (PySide and PyQt4), either via their explictly set preference, via import, or via a default fallback if installed, all of which will be unsupported come QtPy 2.0.

Resolves #261

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,9 +5,11 @@ on:
   push:
     branches:
     - master
+    - '*.x'
   pull_request:
     branches:
     - master
+    - '*.x'
 
 jobs:
   linux:
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -101,6 +101,9 @@ PYSIDE_API = ['pyside']
 # Names of the expected PySide2 api
 PYSIDE2_API = ['pyside2']
 
+# Names of the legacy APIs that we should warn users about
+LEGACY_APIS = PYQT4_API + PYSIDE_API
+
 # Detecting if a binding was specified by the user
 binding_specified = QT_API in os.environ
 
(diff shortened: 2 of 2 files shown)


CAM-Gerlach:
Looks like the CIs are not running at all. I'll see if I can fix that...

CAM-Gerlach:
The issue is the old CI job, before my improvements, only runs on `master`. I'll modify it here so it triggers as well (so long as it works being in a fork from my repo, which is the thing I had trouble with before, otherwise I'll have to do it in another branch/PR here).

CAM-Gerlach:
(So, incidentally, PR #281 skipped the tests and checks completely)

ccordoba12 (approved):

CAM-Gerlach:
I amended it into the same commit, to hopefully make it a little easier when merging to `master` (since adding the warning and the release branch pattern will need to be dropped when merging; `LEGACY_APIS` can stay.

CAM-Gerlach:
@ccordoba12 Want me to merge when the CIs pass, or should we wait for @dalthviz ?

dalthviz:
Is there any need to merge `1.x` with master after this PR @CAM-Gerlach ? In the merge we will need to readd `PyQT4_API` and `PYSIDE_API` or change `LEGACY_APIS` to be `['pyqt', 'pyqt4', 'pyside']` while the rest of the changes are dropped, right?

ccordoba12:
@dalthviz, you can still merge with `master` and drop all changes present here (because they don't apply) by doing an empty commit for the merge.

CAM-Gerlach:
Ah, I somehow forgot that `LEGACY_APIS` also required having those other constants, which probably aren't harmful but realistically aren't really needed, so I guess a just doing a null merge is fine (to still keep the branches in sync of course). Thanks for taking care of it, @dalthviz .

PR #287, opened 2021-11-26
PR: Map exec_ to their non-deprecated alternatives

### LINKED ISSUE #286: PySide6 DeprecationWarning of exec_
Hi, 

exec_ function is deprecated in PySide6, so the common PyQt5/PySide2 code gets deprecated warning when QT_API=pyside6.

```
DeprecationWarning: 'exec_' will be removed in the future. Use 'exec' instead.
  sys.exit(app.exec_())
```

On the other hand, PyQt6 removes exec_ but QtPy provides exec_ for the compatibility. (#259)
As a result, DeprecationWarning of exec_ occurs only when QT_API=pyside6.
Which should this DeprecationWarning be addressed by QtPy or the user application?
Note that print_ function of PySide6 may have the same problem.

kumattau:
This PR fixes #286.

All `exec_`s of PySide6 are simply mapped (assigned) to `exec`s to avoid the following DeprecationWarning of PySide6.

```
DeprecationWarning: 'exec_' will be removed in the future. Use 'exec' instead.
  sys.exit(app.exec_())
```

I think if there is no problem if `exec_` is overwritten, this PR can be merged.
If a way to access the original `exec_` is required, QtPy may provides the option for controlling the overwrite or the special attribute for accessing original `exec_`.

### DIFF
--- a/qtpy/QtCore.py
+++ b/qtpy/QtCore.py
@@ -61,6 +61,12 @@ elif PYSIDE6:
     Qt.TextColorRole = Qt.ForegroundRole
     Qt.MidButton = Qt.MiddleButton
 
+    # Map DeprecationWarning methods
+    QCoreApplication.exec_ = QCoreApplication.exec
+    QEventLoop.exec_ = QEventLoop.exec
+    QThread.exec_ = QThread.exec
+    QTextStreamManipulator.exec_ = QTextStreamManipulator.exec
+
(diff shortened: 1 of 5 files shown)


dalthviz (approved):
Thanks @kumattau for all the help with this one! LGTM :+1:

CAM-Gerlach:
Thanks @kumattau !

PR #288, opened 2021-11-28
QRegExp becomes  QRegularExpression in PySide6

stonebig:
(used per guidata)

### DIFF
--- a/qtpy/QtCore.py
+++ b/qtpy/QtCore.py
@@ -60,7 +60,9 @@ elif PYSIDE6:
     Qt.BackgroundColorRole = Qt.BackgroundRole
     Qt.TextColorRole = Qt.ForegroundRole
     Qt.MidButton = Qt.MiddleButton
+    QRegExp = QRegularExpression
 
+    
     # Map DeprecationWarning methods
     QCoreApplication.exec_ = QCoreApplication.exec
     QEventLoop.exec_ = QEventLoop.exec


CAM-Gerlach (dismissed):
LGTM, thanks @stonebig !

CAM-Gerlach:
Related #233

CAM-Gerlach:
@stonebig Browsing winpython/winpython#1030 I see you mentioned the fix for ``guidata``; will you be sending a PR for that as well?

Also, definitely looking forward to @impact27 's significant performance optimizations in 5.2.1; they only barely missed the cut for 5.2.0.

stonebig:
To get guidata truly working on python-3.10 on pyside6-6.2.1 , I needed that simple fix and re-adding py3compat.py file.

Guidata has not yet dropped python-2.7, so this temporary gymnastic.

kumattau:
QRegularExpression is not compatible with QRegExp.

https://doc.qt.io/qt-5/qregularexpression.html#notes-for-qregexp-users

I think this mapping may cause a surprised error when a user application which use QRegExp runs with QT_API=pyside6.

CAM-Gerlach:
Thanks very much, @kumattau . @stonebig , given per the above docs QRegularExpression is supported in Qt5 (and Qt6), since QtPy 2.x will only officially support PyQt >=5.9 and PySide2 >= 5.12, if those versions support QRegularExpression, would it make sense to just support that with QtPy and update your dependent code accordingly (and maybe raise a `DeprecationWarning`/`PythonQtWarning` if using the deprecated methods, to nudge users to switch to ones compatible with all versions of Qt supported by their current QtPy version?)

dalthviz:
Hi @stonebig , checking this seems like `guidata` imports `QRegExp` but it doesn't use it. Here the line:

https://github.com/PierreRaybaut/guidata/blob/76e4bfdc08991b293ad32ec2786f7a7fbdfa9955/guidata/widgets/codeeditor.py#L58

Maybe then the `QRegExp` import could be removed and in that way prevent the error with QtPy 2.0.0.dev0? I'm not totally sure what could we do from the QtPy side but if you have any idea or suggestion let us know!

Also, thanks @kumattau and @CAM-Gerlach for taking the time to review this :+1:

stonebig:
A yes, it seems just imported...

PR #289, opened 2021-11-29
PR: Add warning for deprecated/EoL Qt5 & PyQt5/PySide2 versions to fix #284

### LINKED ISSUE #284: Warn if using a deprecated and/or unsupported Qt 5 version
Since #261 was implemented in PR #283 , we're now issuing a `DeprecationWarning` for QtPy 1.x when used with Qt versions and bindings that are deprecated, untested in QtPy 1.x, and unsupported in QtPy 2.0.

Given Qt/PyQt versions <5.9 and Pyside2 versions <5.12 are equally deprecated, untested in QtPy 1.x and officially unsupported in QtPy 2.0 (since we've removed any compat code for them, AFAIK), it would seem to make sense to issue an equivalent `DeprecationWarning` in QtPy 1.x for that case as well.

Furthermore, we might want to elevate the warning to a `UserWarning` or a `PythonQtError` in QtPy 2.0, given that serious runtime breakage may occur, just as we currently do in 2.0 for the Qt 4 bindings, so users see it up front rather than having things mysteriously break later.

Depending on what is decided in #233 or elsewhere for future QtPy 2.0 versions going forward, we could also consider introducing a DeprecationWarning in QtPy 2.0 for PyQt <5.12 in that version, if in #233 it is decided to that be our support target for future follow-on versions. However, if that's not immediately decided or out of scope here, that could be punted again to another followup issue.

Relevant conversation history on #261:

<details>

I said

> We might want to consider a warning if a version of PyQt (e..g. < 5.9 or < 5.12), and/or PySide2 (e.g. < 5.12) is used that would similarly be unsupported in QtPy 2.0.0, and untested in recent QtPy 1.x.

@dalthviz said

> Thats true, yes, I would say that the acceptable minimum for PyQt could be >=5.9 (since is the lates version in the anaconda channel) but since there are some issues regarding scoped enums with that version for Python 3.6 and 3.7 probably at the end the minimum will be something like >=5.12 as with PySide2. However we need to further discuss this too

I said:

> I assumed this issue was motivated by the fact that Qt4 binding support is removed in QtPy 2.0, untested and deprecated in QtPy 1.x, and EoL upstream. Since all three of these reasons are equally true for Qt5 <5.9/5.12 (the specifics of which are being discussed in #233 ), I thought it might make sense to consider warning for that here as well (considering I'd think we'd want to decide on that before you make a final QtPy 1.x release with this change).

@dalthviz said

> Ah yes that makes sense, will reword this issue then :+1: 

I said

> @ccordoba12 Oh, so we already warn on Qt <5.6 and/or <5.9?

@dalthviz said

> Also @ccordoba12 you think that in the message we should add something regarding Qt5 ? (Like a minimum version)

I said

> point_up Since Qt bindings < 5.9-5.12 (per the discussion on #233 ) are also equally deprecated and no longer supported in QtPy 2.0, my suggestion was that we should also include those in this warning too.

@ccordoba12 said

> I don't think so because qtpy takes care of that for the user.

I said

> @ccordoba12 Oh, so we already warn on Qt <5.6 and/or <5.9?

@ccordoba12 said

> No, we don't. Perhaps we should warn about Qt < 5.9 because the oldest version we can test.

@dalthviz said

> Checking again the comments, regarding the message for Qt 5 <5.9, that's going to be done by @CAM-Gerlach in a new PR, right? 

@ccordoba12 said

> I think we should leave this message for QtPy 2.0

I said

> I'm not sure about that, but we can discuss that in the new issue. 

_Originally posted by @CAM-Gerlach in https://github.com/spyder-ide/qtpy/issues/261#issuecomment-952260141_

CAM-Gerlach:
As discussed in #284 , adds a `DeprecationWarning` on QtPy 1.x for Qt5/PyQt5 versions below 5.9 and PySide2 versions below 5.12.

@dalthviz the merge to 1.x will be rather non-trivial (replacing distutils with packaging, changing the warning type, fixing merge conflicts, and possibly a tweak to the message to include users as well, etc) so I'm happy to take care of that too, if you'd like.

Closes #284

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -104,6 +104,11 @@ PYSIDE2_API = ['pyside2']
 # Names of the legacy APIs that we should warn users about
 LEGACY_APIS = PYQT4_API + PYSIDE_API
 
+# Minimum fully supported versions of Qt and the bindings
+PYQT_VERSION_MIN = '5.9.0'
+PYSIDE_VERSION_MIN = '5.12.0'
+QT_VERSION_MIN = '5.9.0'
+
 # Detecting if a binding was specified by the user
 binding_specified = QT_API in os.environ
 
@@ -241,6 +246,17 @@ try:
 except (ImportError, PythonQtError):
     pass
 
+
+def _warn_old_minor_version(name, old_version, min_version):
+    warning_message = (
+        "{name} version {old_version} is unsupported upstream and "
+        "deprecated by QtPy. To ensure your application is still supported "
+        "in QtPy 2.0, please make sure it doesn't depend upon {name} versions "
+        "older than {min_version}.".format(
+            name=name, old_version=old_version, min_version=min_version))
+    warnings.warn(warning_message, DeprecationWarning)
+
+
 # Warn if using a legacy, soon to be unsupported Qt API/binding
 if API in LEGACY_APIS or initial_api in LEGACY_APIS:
     warnings.warn(
@@ -251,3 +267,12 @@ if API in LEGACY_APIS or initial_api in LEGACY_APIS:
         "set the 'QT_API' env var to 'pyqt', 'pyqt4' or 'pyside'.",
         DeprecationWarning,
     )
+else:
+    if LooseVersion(QT_VERSION) < LooseVersion(QT_VERSION_MIN):
+        _warn_old_minor_version('Qt', QT_VERSION, QT_VERSION_MIN)
+    if PYQT_VERSION and (LooseVersion(PYQT_VERSION)
+                         < LooseVersion(PYQT_VERSION_MIN)):
+        _warn_old_minor_version('PyQt', PYQT_VERSION, PYQT_VERSION_MIN)
+    elif PYSIDE_VERSION and (LooseVersion(PYSIDE_VERSION)
+                             < LooseVersion(PYSIDE_VERSION_MIN)):
+        _warn_old_minor_version('PySide', PYSIDE_VERSION, PYSIDE_VERSION_MIN)


qtpy/__init__.py
```diff
@@ -241,6 +241,19 @@ class PythonQtWarning(Warning):
 except (ImportError, PythonQtError):
     pass
 
+PYQT_VERSION_MIN = '5.9.0'
+PYSIDE_VERSION_MIN = '5.12.0'
+QT_VERSION_MIN = '5.9.0'
```
dalthviz, line 246: Maybe this constants should go above? Near the `LEGACY_APIS` constant

qtpy/__init__.py
```diff
@@ -251,3 +264,12 @@ class PythonQtWarning(Warning):
         "set the 'QT_API' env var to 'pyqt', 'pyqt4' or 'pyside'.",
         DeprecationWarning,
     )
+else:
+    if LooseVersion(QT_VERSION) < LooseVersion(QT_VERSION_MIN):
+        _warn_old_minor_version('Qt', QT_VERSION, QT_VERSION_MIN)
+    if PYQT_VERSION and (LooseVersion(PYQT_VERSION)
+                         < LooseVersion(PYQT_VERSION_MIN)):
+        _warn_old_minor_version('PyQt', PYQT_VERSION, PYQT_VERSION_MIN)
```
dalthviz, line 272: Maybe we should put here `PyQt5`?

qtpy/__init__.py
```diff
@@ -251,3 +264,12 @@ class PythonQtWarning(Warning):
...
+    elif PYSIDE_VERSION and (LooseVersion(PYSIDE_VERSION)
+                             < LooseVersion(PYSIDE_VERSION_MIN)):
+        _warn_old_minor_version('PySide', PYSIDE_VERSION, PYSIDE_VERSION_MIN)
```
dalthviz, line 275: Maybe we should put here `PySide2`?

dalthviz (commented):
Thanks @CAM-Gerlach for the help with this! I left some questions/suggestions. Regarding the merge with master, sounds go to me if do it (after merging this then you can go ahead with that :+1:)

ccordoba12 (approved):
Looks good to me, thanks @CAM-Gerlach for your work on this!

qtpy/__init__.py
CAM-Gerlach: I actually thought about that, but I made it `PyQt` to avoid repeating the major version twice (`PyQt5 5.9`), to match the constants and so that we wouldn't need to change it if we check for a minimum PyQt6 version. But I can change it if you're sure.

qtpy/__init__.py
CAM-Gerlach: Maybe, see above ^

qtpy/__init__.py
CAM-Gerlach: Yeah, I was going to move them but forgot, thanks. Doing that now.

CAM-Gerlach: Done, also added a comment and added a second line break around the new private function

qtpy/__init__.py
dalthviz: That makes sense, lets leave it as it is then. Thanks for claryfing @CAM-Gerlach :+1:

dalthviz (approved):
Thanks @CAM-Gerlach for your help with this!

CAM-Gerlach:
Thanks @dalthviz and @ccordoba12 for the reviews and feedback! Taking care of the `master` merge now.

CAM-Gerlach:
Seems like there's some weird behavior with Win 10 + PyQt 5.9 + conda + Python 3.6, where the tests pass, but when we run the command to save the path for the test coverage, which imports QtPy a _second_ time, `QT_VERSION` is not defined, which means that all the `if` blocks fell through or `ImportError`ed. I opened PR #292 which fixes the proximate and secondary problems, and will investigate locally to see if I can reproduce what could point to another underlying issue.

PR #292, opened 2021-12-01
PR: Define Qt/binding versions at top level, fix warnings if versions not found, and fix test dir on CIs

CAM-Gerlach:
Followup to merging PR #289 to `master` following a very weird and unexpected check failure on Win 10 + PyQt 5.9 + conda + Python 3.6; see [CI logs](https://github.com/spyder-ide/qtpy/pull/292#issue-793185869).

Seems like there's some weird behavior on this very specific CI run, where the tests pass, but when we run the command to save the path for the test coverage, which imports QtPy a _second_ time in a new instance of the same interpreter, `QT_VERSION` is not defined, which must mean that all the `if` blocks either didn't trigger or had an `ImportError` and QtPy fell back to using PyQt5. I could understand if PyQt 5.9 didn't define those top-level constants (since the source repo is not public, I tried to check beforehand but couldn't), but the fact that it worked without error the first time for the tests leads me to believe that it must either have been something with setting the env variables that triggered the different behavior on the second, or there's something else I'm not accounting for going on.

EDIT: It was the latter, the second Python command was run without `-I` which resulted in shadowing due to the hack I used to avoid problems due to the non-src directory layout. See below for full details.

~While this PR fixes the proximate error and ensures that the top-level constants are always at least defined, I want to see if I can reproduce this locally in the same env, as it may indicate an underlying issue.~ Done, see below.

So, this PR implements three levels of fixes to the related issues discovered here:

- [x] Checks that `QT_VERSION` is defined first before acting on it (like all the others)
- [x] Initializes the top-level version constants to `None` at the top of the file, which avoids them being undefined if no Qt API is found and also reduces duplication and verbosity
- [x] Fixes the shadowing issue in the tests that surfaced this behavior to begin with

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -105,6 +105,6 @@ jobs:
           COVERALLS_FLAG_NAME: ${{ matrix.os }} Python ${{ matrix.python-version }} conda=${{ matrix.use-conda }}
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          cd qtpy  # Switch to test working dir per non-src-layout hack
+          cd temp_test_dir  # Switch to test working dir per non-src-layout hack
           cat qtpy_basedir.txt
           pipx run coveralls --service=github --rcfile="../.coveragerc" --basedir="$(cat qtpy_basedir.txt)"
--- a/.github/workflows/test.sh
+++ b/.github/workflows/test.sh
@@ -53,13 +53,14 @@ echo dist/*.whl | xargs -I % python -bb -X dev -W error -m pip install --upgrade
 conda list
 
 # Run tests
-cd qtpy  # Hack to work around non-src layout pulling in local instead of installed package for cov
-python -I -bb -X dev -W error -m pytest --cov-config ../.coveragerc --cov-append
+mkdir -p temp_test_dir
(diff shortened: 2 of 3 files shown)


CAM-Gerlach:
Okay, I figured out what's going on. Since QtPy doesn't use the recommended `src` directory layout, I had to `cd` into another directory, in this case `qtpy`, to avoid testing the local copy rather than the actual installed package (which not only would not detect any packaging problems, but also plays havoc with the coverage). However, this was a rather short-sighted fix, as this results in QtPy's `sip` module shadowing the `sip` package (and could result in other such tricky issues). This doesn't cause problems for the tests since the `importlib` import mode prevents the shadowing from occurring, but this does mean that when the standalone command is run to import qtpy and get its path for the coverage, importing `PyQt5` fails due to this and thus version is left undefined. I'll fix that as well here, to avoid any future knock-on effects from this, by just cding into a temporary directory instead.

CAM-Gerlach:
@dalthviz @ccordoba12 if either of you could look this over and merge this if there's nothing serious going on, so that tests are passing on `master` again, that would be great, thanks.

ccordoba12 (approved):

CAM-Gerlach:
Thanks @ccordoba12 !

PR #294, opened 2021-12-03
PR: Add minimum version check for Qt6 >=6.2.0

### LINKED ISSUE #233: PyQt6 support

CAM-Gerlach:
Per the discussion in #233 , adds a minimum version check for Qt6 and each binding to ensure they are at least version 6.2.0 or higher, since we don't officially support anything under that. Also, refactors the constants a bit accordingly.

Closes #233 (again)

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -106,8 +106,8 @@ initial_api = API
 assert API in (PYQT5_API + PYQT6_API + PYSIDE2_API + PYSIDE6_API)
 
 is_old_pyqt = is_pyqt46 = False
-PYQT5 = True
-PYQT4 = PYQT6 = PYSIDE = PYSIDE2 = PYSIDE6 = False
+QT5 = PYQT5 = True
+QT4 = QT6 = PYQT4 = PYQT6 = PYSIDE = PYSIDE2 = PYSIDE6 = False
 
 PYQT_VERSION = None
 PYSIDE_VERSION = None
@@ -128,6 +128,8 @@ if API in PYQT5_API:
     try:
         from PyQt5.QtCore import PYQT_VERSION_STR as PYQT_VERSION  # analysis:ignore
         from PyQt5.QtCore import QT_VERSION_STR as QT_VERSION  # analysis:ignore
+
+        QT5 = PYQT5 = True
 
         if sys.platform == 'darwin':
             macos_version = parse(platform.mac_ver()[0])
@@ -152,8 +154,10 @@ if API in PYQT6_API:
     try:
         from PyQt6.QtCore import PYQT_VERSION_STR as PYQT_VERSION  # analysis:ignore
         from PyQt6.QtCore import QT_VERSION_STR as QT_VERSION  # analysis:ignore
-        PYQT5 = False
-        PYQT6 = True
+
+        QT5 = PYQT5 = False
+        QT6 = PYQT6 = True
+
     except ImportError:
         API = os.environ['QT_API'] = 'pyside2'
 
@@ -164,7 +168,7 @@ if API in PYSIDE2_API:
         from PySide2.QtCore import __version__ as QT_VERSION  # analysis:ignore
 
         PYQT5 = False
-        PYSIDE2 = True
+        QT5 = PYSIDE2 = True
 
         if sys.platform == 'darwin':
             macos_version = parse(platform.mac_ver()[0])
@@ -184,8 +188,8 @@ if API in PYSIDE6_API:
         from PySide6 import __version__ as PYSIDE_VERSION  # analysis:ignore
         from PySide6.QtCore import __version__ as QT_VERSION  # analysis:ignore
 
-        PYQT5 = False
-        PYSIDE6 = True
+        QT5 = PYQT5 = False
+        QT6 = PYSIDE6 = True
 
     except ImportError:
         API = os.environ['QT_API'] = 'pyqt5'
@@ -203,7 +207,7 @@ API_NAME = {'pyqt6': 'PyQt6', 'pyqt5': 'PyQt5',
 try:
     # QtDataVisualization backward compatibility (QtDataVisualization vs. QtDatavisualization)
     # Only available for Qt5 bindings > 5.9 on Windows
-    from . import QtDataVisualization as QtDatavisualization
+    from . import QtDataVisualization as QtDatavisualization  # analysis:ignore
 except (ImportError, PythonQtError):
     pass
 
@@ -211,8 +215,8 @@ except (ImportError, PythonQtError):
 def _warn_old_minor_version(name, old_version, min_version):
     """Warn if using a Qt or binding version no longer supported by QtPy."""
     warning_message = (
-        "{name} version {old_version} is not supported upstream or "
-        "by QtPy 2.0. To ensure your application works correctly with QtPy, "
+        "{name} version {old_version} is not supported by QtPy. "
+        "To ensure your application works correctly with QtPy, "
         "please upgrade to {name} {min_version} or later.".format(
             name=name, old_version=old_version, min_version=min_version))
     warnings.warn(warning_message, PythonQtWarning)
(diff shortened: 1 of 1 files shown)


ccordoba12 (approved):

kumattau:
@CAM-Gerlach, @ccordoba12,

How about introducing PYQT5_VERSION, PYSIDE2_VERSION, PYQT6_VERSION and PYSIDE6_VERSION ?

l think, 

`parse(PYQT5_VERSION) <　parse(PYQT5_VERSION_MIN)`

is better than

`parse(PYQT_VERSION) < parse(PYQT5_VERSION_MIN)`

and others are the same.

CAM-Gerlach:
Hmm, thanks for the suggestion @kumattau . My take is that it might look a little better in a few parts of the warning check code, but I'm not sure that a bit of internal-only aesthetics is worth introducing even more new package-top-level public constants (that would have to be changed with QtPy versions that added/removed binding support) which are rather redundant to each other (since only one is going to be defined at a time) and the need to define and check them would add further code complexity to a (difficult to reliably unit test) warnings check. Additionally, on the aesthetics front, adding that one addition character would push several of the lines over the 79 char limit, making it even longer, less consistent and harder to read at a glance.

@dalthviz , what do you think?

kumattau:
Let me explain the background of my proposal in a little more detail.
First, PYQT_VERSION and PYSIDE_VERSION are expected to be public APIs for users because these don't start with an underscore.

In the current API, the following simply comparison

`parse(PYQT_VERSION) < parse(PYQT6_VERSION_MIN)`

is incorrect usage, but it is difficult for users to understand this.

The related public APIs is as follows.

* QT_API environment variable
  pyqt5, pyside2, pyqt6 or pyside6 is set

* PYQT5, PYSIDE2, PYSIDE2, PYSIDE6
  corresponding the above QT_API, the only one of these constants is set to True, others are set to False

By introducing PYQT5_VERSION, PYSIDE2_VERSION, PYQT6_VERSION and PYSIDE6_VERSION, XXX_VERSIONs become the same behavior as  PYQT5, PYSIDE2, PYSIDE2, PYSIDE6 and users can use XXX_VERSIONs without mistake.

Additionaly, I think introducing QT5_VERSION and QT6_VERSION is better from the API consistency perspective.

The above is my personal opinion, and of course, I respect the judgment of the spyder team.

qtpy/__init__.py
```diff
@@ -207,17 +211,26 @@ class PythonQtWarning(Warning):
 def _warn_old_minor_version(name, old_version, min_version):
     """Warn if using a Qt or binding version no longer supported by QtPy."""
     warning_message = (
-        "{name} version {old_version} is no longer supported upstream or "
+        "{name} version {old_version} is not supported upstream or "
         "by QtPy 2.0. To ensure your application works correctly with QtPy, "
```
dalthviz, line 215: Maybe we can simplify the warning message like this?
```suggestion
        "{name} version {old_version} is not supported by QtPy. "
        "To ensure your application works correctly with QtPy, "
```

qtpy/__init__.py
```diff
@@ -207,17 +211,26 @@ class PythonQtWarning(Warning):
...
-    _warn_old_minor_version('Qt', QT_VERSION, QT_VERSION_MIN)
-if PYQT_VERSION and parse(PYQT_VERSION) < parse(PYQT_VERSION_MIN):
-    _warn_old_minor_version('PyQt', PYQT_VERSION, PYQT_VERSION_MIN)
-elif PYSIDE_VERSION and parse(PYSIDE_VERSION) < parse(PYSIDE_VERSION_MIN):
-    _warn_old_minor_version('PySide', PYSIDE_VERSION, PYSIDE_VERSION_MIN)
+# Warn if using an End of Life or unsupported Qt API/binding minor version
+if QT_VERSION:
+    if parse(QT_VERSION) < parse(QT5_VERSION_MIN):
+        _warn_old_minor_version('Qt5', QT_VERSION, QT5_VERSION_MIN)
+    elif parse('6') <= parse(QT_VERSION) < parse(QT6_VERSION_MIN):
```
dalthviz, line 225: Checking @kumattau comments this vaidation looks a little bit confusing at first glance.

Maybe we could add some boolean constants like `QT5` and `QT6` (following the same spirit as the existing ones (`PYQT5`, `PYQT6`, `PYSIDE2` and `PYSIDE6`) where ony one is `True` at a time. I think that could help making the validation more clear, something like:

```python

elif QT6 and parse(QT_VERSION) < parse(QT6_VERSION_MIN):

```

 What do you think @CAM-Gerlach @kumattau ? Is adding this new constants a good compromise between clarity vs new constants/redundancy?

dalthviz (commented):
Thanks @CAM-Gerlach for working on this! I left some comments/suggestions/questions to:

* Simplify the warning message 
* Add more carity to the validation on Qt6 taking into account the comments of @kumattau (thanks for taking the time to check this and raising your concerns by the way :+1:)

qtpy/__init__.py
kumattau: I think @dalthviz 's idea is better from the perspective of consistency.

But since it is difficult to understand the different ranges represented by each constant,

* PYQT5 : True if PyQt5 is selected
* PYQT_VERSION : PYQT5 or PYQT6 version
* PYQT5_VERSION_MIN : minimum version of supported PyQt5
* PYQT5_VERSION is nothing.

(PYSIDE and QT are the same as PYQT)

If these constants are public API for users but PYQT5_VERSION and so on are not added,
I think it is better approach that the usage of related constants is explained clearly in the document like as follows:

```python
# check Qt6 version (need to check QT6 first)
if QT6 and parse(QT_VERSION) < parse(QT6_VERSION_MIN):
    print("cannot support or deprecated")

# check Qt5 version (need to check QT5 first)
if QT5 and parse(QT_VERSION) < parse(QT5_VERSION_MIN):
    print("cannot support  or deprecated")

# check PySide2 version (need to check PYSIDE2 first)
if PYSIDE2 and parse(PYSIDE_VERSION) < parse(PYSIDE2_VERSION_MIN)
    print("cannot support  or deprecated")

# check PySide6 version (need to check PYSIDE6 first)
if PYSIDE6 and parse(PYSIDE_VERSION) < parse(PYSIDE2_VERSION_MIN)
    print("cannot support  or deprecated")

# check PyQt5 version (need to check PYQT5 first)
if PYQT5 and parse(PYQT_VERSION) < parse(PYQT5_VERSION_MIN)
    print("cannot support  or deprecated")

# check PyQt6 version (need to check PYQT6 first)
if PYQT6 and parse(PYQT_VERSION) < parse(PYQT6_VERSION_MIN)
    print("cannot support  or deprecated")
```

qtpy/__init__.py
CAM-Gerlach: Agreed; I almost changed to something resembling that, but just didn't want to stray too far from the message we'd agreed on previously.

qtpy/__init__.py
CAM-Gerlach: I agree with @dalthviz 's approach as well, would be useful to have those top-level constants and makes the code much cleaner while avoiding the extra complexity of the previous suggested approach. I also agree with explicitly checking the constants in each branch, for clarity and completeness, though the existence of the version constant has to be checked as well at least with the current `__init__` structure, as that is imported from the Qt bindings as well and may not be defined if all of the imports fail and it falls back to the default, causing a hard crash. I also separated the Qt and bindings checks by a line, to make the structure more clear. 

Here's what the final result looks like, which I've pushed here:

```python
# Warn if using an End of Life or unsupported Qt API/binding minor version
if QT_VERSION:
    if QT5 and (parse(QT_VERSION) < parse(QT5_VERSION_MIN)):
        _warn_old_minor_version('Qt5', QT_VERSION, QT5_VERSION_MIN)
    elif QT6 and (parse(QT_VERSION) < parse(QT6_VERSION_MIN)):
        _warn_old_minor_version('Qt6', QT_VERSION, QT6_VERSION_MIN)

if PYQT_VERSION:
    if PYQT5 and (parse(PYQT_VERSION) < parse(PYQT5_VERSION_MIN)):
        _warn_old_minor_version('PyQt5', PYQT_VERSION, PYQT5_VERSION_MIN)
    elif PYQT6 and (parse(PYQT_VERSION) < parse(PYQT6_VERSION_MIN)):
        _warn_old_minor_version('PyQt6', PYQT_VERSION, PYQT6_VERSION_MIN)
elif PYSIDE_VERSION:
    if PYSIDE2 and (parse(PYSIDE_VERSION) < parse(PYSIDE2_VERSION_MIN)):
        _warn_old_minor_version('PySide2', PYSIDE_VERSION, PYSIDE2_VERSION_MIN)
    elif PYSIDE6 and (parse(PYSIDE_VERSION) < parse(PYSIDE6_VERSION_MIN)):
        _warn_old_minor_version('PySide6', PYSIDE_VERSION, PYSIDE6_VERSION_MIN)
```

dalthviz (approved):
Thanks @CAM-Gerlach for working on this! LGTM :+1:

Also, thanks @kumattau for the feedback! :+1:

Issue #269, opened 2021-10-30
Add CI testing and official support for Python 3.10


CAM-Gerlach:
Originally, part of #253 included adding CI testing, and thus official support for Python 3.10, but it turned out in PR #262 that given the conda packages for Pytest aren't officially yet declared compatible (despite the upstream PyPI packages having the 3.10 tag), it would more complex/inefficient than anticipated to add it right away, and so it was deferred to a followup.

As of now, we can either wait (hopefully <= a couple weeks) for the conda Pytest, etc packages to officially add Py3.10 support (simplest and least CI impact, but not immediate), or install the test deps (pytest and pytest-cov, etc) using separate conda/pip commands, using Python 3.10 on conda=no builds (increased CI complexity and runtime, especially with the matrix simplification and the coverage improvements, though allows us to add support right away).

I presume we want to target this for the QtPy 2.0.0 release, so while the former would be preferable, I imagine we'd want to go with the latter if it is not possible in time. @dalthviz , would it be helpful for us to open an issue on the pytest conda feedstock repo about this, and/or somewhere else? I didn't see one last time I checked.

CAM-Gerlach:
Seems like the "waiting" approach is blocked on the cpython3.10 migration for CF, which per the discussion in conda-forge/conda-forge-pinning-feedstock#2008 may be 1-2 months at least.

So to ensure this doesn't block QtPy v2.0.0, I guess I'll just go ahead with the approach of bifurcating the `conda=yes` and `conda=no` branches in `test.sh` a bit more so the latter uses a pure `pip` test dep stack and thus can run under Python 3.10. with a bit of bash-fu in the `test.sh` I should be able to avoid appreciable CI run time impacts on either the conda=yes or no installs, and using `exclude` _should_ allow doing so with minimal extra verbosity in the matrix (unless unfixable Py3.10 problems arise), so there's little real downside to just going ahead with this, aside from a modest amount of extra time up front.

CAM-Gerlach:
@dalthviz This is pretty much blocked on PR #268 being merged, as significant merge conflicts will result, though that PR is pretty straightforward and umabigous so I don't expect too much delay on that. It should be completely orthogonal to #270 getting addressed, minus any unforeseen issues.

dalthviz:
Thinking about this maybe we should just wait for a bit

CAM-Gerlach:
[...]

PR #295, opened 2021-12-07
PR: Drop support for Python 3.6

ccordoba12:
That's because that version is about to reach end-of-life.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,17 +37,17 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ['3.6', '3.9']
+        python-version: ['3.7', '3.9']
         use-conda: ['Yes', 'No']
         include:
         - os: ubuntu-latest
           special-invocation: 'xvfb-run --auto-servernum '
-        - python-version: '3.6'
+        - python-version: '3.7'
           use-conda: 'No'
-          skip-pip-check: 'true'  # Pytest packaging issue on Python 3.6 defaults channel
+          skip-pip-check: 'true'  # Pytest packaging issue on Python 3.7 defaults channel
           pyside2-version: 5.12.0  # 5.12.1-5.12.6 fails on collection/segfaults on patch test
         - os: ubuntu-latest
-          python-version: '3.6'
+          python-version: '3.7'
           use-conda: 'No'
           skip-pyqt6: true  # No wheels on Py 3.6 Linux CIs
         - os: windows-latest
@@ -55,13 +55,14 @@ jobs:
           use-conda: 'No'
           pyside2-version: 5.15  # No 5.12 wheel on Windows and Python 3.9
         - os: windows-latest
-          python-version: '3.6'
-          use-conda: 'Yes'
-          pyqt5-qt-version: '5.9'  # 5.12 is apparently unreliable here
-        - os: macos-latest
-          python-version: '3.6'
+          python-version: '3.7'
           use-conda: 'No'
-          skip-pyqt6: true  # No wheels on Py 3.6 macOS CIs
+          skip-pip-check: 'true'  # Pytest packaging issue on Python 3.7 defaults channel
+        - os: macos-latest
+          python-version: '3.7'
+          use-conda: 'No'
+          skip-pyqt6: true  # No wheels on Py 3.7 macOS CIs
+          skip-pip-check: 'true'  # Pytest packaging issue on Python 3.7 defaults channel
     steps:
       - name: Checkout branch
         uses: actions/checkout@v2
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,7 +1,7 @@
 [metadata]
 name = QtPy
 version = attr: qtpy.__version__
-description = Provides an abstraction layer on top of the various Qt bindings (PyQt5/6 and PySide2/6) and additional custom QWidgets.
+description = Provides an abstraction layer on top of the various Qt bindings (PyQt5/6 and PySide2/6)
 long_description = file: README.md
 long_description_content_type = text/markdown
 url = https://github.com/spyder-ide/qtpy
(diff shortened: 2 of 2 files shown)


setup.cfg
```diff
@@ -24,10 +24,10 @@ classifiers =
     Operating System :: OS Independent
     Programming Language :: Python :: 3
     Programming Language :: Python :: 3 :: Only
-    Programming Language :: Python :: 3.6
     Programming Language :: Python :: 3.7
     Programming Language :: Python :: 3.8
     Programming Language :: Python :: 3.9
+    Programming Language :: Python :: 3.10
```
dalthviz, line 30: Should we add support for Python 3.10 in this PR then (and maybe associate this PR to https://github.com/spyder-ide/qtpy/issues/269)?

dalthviz (commented):
Thanks @ccordoba12 ! I left a comment/question regarding the Python 3.10 classifier addition. Otherwise LGTM :+1:

setup.cfg
ccordoba12: I don't have time for that, but I doubt QtPy 2 will have trouble working with Python 3.10 because we're not using specific languages features that will prevent it.

CAM-Gerlach: Like @dalthviz , I don't understand why we would formally declare support for Python 3.10 before actually testing it, and this is out of scope for this PR, given this is the declared purpose of issue #269 , which I'll take care of one way or another before the final QtPy 2.0 release.

.github/workflows/ci.yml
```diff
@@ -37,31 +37,32 @@ jobs:
...
+        python-version: ['3.7', '3.9']
         use-conda: ['Yes', 'No']
         include:
         - os: ubuntu-latest
           special-invocation: 'xvfb-run --auto-servernum '
-        - python-version: '3.6'
+        - python-version: '3.7'
           use-conda: 'No'
-          skip-pip-check: 'true'  # Pytest packaging issue on Python 3.6 defaults channel
+          skip-pip-check: 'true'  # Pytest packaging issue on Python 3.7 defaults channel
```
CAM-Gerlach, line 47: Did you verify this skip was still required?

.github/workflows/ci.yml
```diff
@@ -37,31 +37,32 @@ jobs:
...
           pyside2-version: 5.12.0  # 5.12.1-5.12.6 fails on collection/segfaults on patch test
         - os: ubuntu-latest
-          python-version: '3.6'
+          python-version: '3.7'
           use-conda: 'No'
           skip-pyqt6: true  # No wheels on Py 3.6 Linux CIs
```
CAM-Gerlach, line 52: This skip (and thus this whole section) can likely can be removed, unless the issue also persists on the 3.7 wheels

.github/workflows/ci.yml
```diff
@@ -37,31 +37,32 @@ jobs:
...
           python-version: '3.9'
           use-conda: 'No'
           pyside2-version: 5.15  # No 5.12 wheel on Windows and Python 3.9
         - os: windows-latest
-          python-version: '3.6'
-          use-conda: 'Yes'
-          pyqt5-qt-version: '5.9'  # 5.12 is apparently unreliable here
+          python-version: '3.7'
+          use-conda: 'No'
+          skip-pip-check: 'true'  # Pytest packaging issue on Python 3.7 defaults channel
```
CAM-Gerlach, line 60: I'm not sure I understand this change. This means we're no longer testing PyQt 5.9 on any CI now, despite still declaring compatibility with it (and maintaining such all the way up until the release), and it being the latest version available on the `defaults` channel. Furthermore, I'm not sure why the `skip-pip-check` was repeated here (assuming you verified it is still necessary at all), since it was already specified for all `3.7` runners above.

.github/workflows/ci.yml
```diff
@@ -37,31 +37,32 @@ jobs:
...
         - os: macos-latest
-          python-version: '3.6'
+          python-version: '3.7'
           use-conda: 'No'
-          skip-pyqt6: true  # No wheels on Py 3.6 macOS CIs
+          skip-pyqt6: true  # No wheels on Py 3.7 macOS CIs
```
CAM-Gerlach, line 64: Did you verify this is still the case?

.github/workflows/ci.yml
```diff
@@ -37,31 +37,32 @@ jobs:
...
+          skip-pip-check: 'true'  # Pytest packaging issue on Python 3.7 defaults channel
```
CAM-Gerlach, line 65: Again, this was already specified above (assuming you verified its necessary at all), so not sure why it is repeated here?

CAM-Gerlach (changes requested):
I'm a strong :-1: on this for the QtPy 2.0 release, because:

* We've tested and maintained compatibility all the way up to the release, so it doesn't really make sense to suddenly drop it now and make it impossible for Python 3.6 users to install, despite it being fully functional
* There isn't a compelling benefit to do so immediately; no compat code was simplified/removed and it didn't simplify the CI setup at all (at least as of the current state of the PR); running `pyupgrade --py37-plus` over the full codebase, only one line was changed, and that one only trivially.
* It means all of our tens of thousands of downstream dependencies will either have to drop Python 3.6 support or not upgrade to QtPy 2.0 until they do, and there will be never be a supported combination of Qt6 and Python 3.6
* All of our supported bindings (Pyside2, Pyside6, PyQt5, PyQt6) all still support Python 3.6, so it would be premature to drop it before any of them do
* Checking the [PyPI download stats](https://pypistats.org/packages/qtpy), besides the large spike for 3.7 on Linux (likely driven by CI/automation systems), over the past six months Python 3.6 was a strong number 3 behind Python 3.8 in popularity for QtPy downloads, occasionally even exceeding it.

Furthermore, if we do decide to merge this, there are several specific additional items that need to be investigated/confirmed, and this blocks work on #269 until it is closed or resolved since it touches the same lines.

I strongly recommend that we close this PR for now, retain the existing Python 3.6 support in QtPy 2.0.0, and then drop it in a future 2.1.0 or a later minor version after the EoL, subject to revision if PyQt/PySide dropping support or a non-trivial Py36-only bug crops up that I'm not able to easily solve. I'm happy to commit to taking care of this myself, including verifying all the skips, at any point in QtPy 2.1.0 or later and after the EoL next month.

dalthviz:
Maybe is better to leave this for a future release. What do you think @ccordoba12 ?

dalthviz:
Thanks @ccordoba12 for the understanding! Will then create an issue to keep track of this for the future then :+1:

PR #296, opened 2021-12-09
PR: Add support for Python 3.10 on CIs and packaging

### LINKED ISSUE #269: Add CI testing and official support for Python 3.10

CAM-Gerlach:
Adds Python 3.10 to the CIs, and official support to the packaging metadata, and takes care of a few related warnings.

Also, simplifies the CI config a bit, unskips the Linux PySide6 Py3.6 wheels since they are now available (the macOS ones now are too, but they are failing on the latest macOS), and skips PySide2 for now on Python 3.10, since it doesn't (yet) support it.

Fixes #269

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,23 +37,24 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: ['3.6', '3.9']
+        python-version: ['3.6', '3.10']
         use-conda: ['Yes', 'No']
         include:
         - os: ubuntu-latest
-          special-invocation: 'xvfb-run --auto-servernum '
+          special-invocation: 'xvfb-run --auto-servernum '  # Needed for GUI tests to work
+        - python-version: '3.10'
+          skip-pyside2: true  # Skip Pyside2 on all Python 3.10 builds until it supports it
+        - use-conda: 'Yes'  # No conda packages yet for Qt6
+          skip-pyqt6: true
+          skip-pyside6: true
         - python-version: '3.6'
           use-conda: 'No'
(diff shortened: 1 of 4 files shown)


CAM-Gerlach:
It looks like Pyside2 still isn't supported on Python 3.10, with no wheels having been built and with the conda build failing per conda-forge/pyside2-feedstock#91 , so I've skipped it on Python 3.10 for now. Seems like [this](https://bugreports.qt.io/browse/PYSIDE-1436) might be the issue, or at least one of them, links to a patch on Fedora. @jschueller any idea when we might see a fix, at least as a patch to the conda builds?

jschueller:
pyside is way outdated on conda-forge because qt is lagging at the moment, so I dont expect it to work very soon

dalthviz (approved):
Thanks @CAM-Gerlach for the help with this! LGTM :+1:

PR #298, opened 2021-12-11
make QLibraryInfo.location works

stonebig:
(found while trying to patch Pyzo for PyQt6-6.2.2)


dalthviz (approved):
Thanks @stonebig ! LGTM :+1:

PR #299, opened 2021-12-11
'width' need for PyQt6

stonebig:
same as PySide6


dalthviz (approved):
Thanks @stonebig for the help! LGTM :+1:

PR #303, opened 2021-12-21
PR: Fix QFileSystemModel for PyQt6

almarklein:
For some reason they moved it to QtGui.

FWIW, I'm doing some CI builds for Pyzo, and the one for Linux/py39/PyQt6 broke on this. After the proposed change, it worked. Also see https://stackoverflow.com/questions/66872811/is-the-icon-functionality-of-qfilesystemmodel-broken-in-pyqt6-on-windows?noredirect=1&lq=1


dalthviz (approved):
Thank you @almarklein for helping with this! LGTM :+1:

PR #313, opened 2022-01-13
PR: Add missing `QtGui` utility function to `QtCore.Qt` for PySide bindings

dalthviz:
Fixes #311 

Edit: Also fixes an issue with PyQt6 and sip to use `mightBeRichText`:

imagen

PR #315, opened 2022-01-17
PR: Add `QtCharts` alias for backward compatibility with 1.x

### LINKED ISSUE #304: Qt6: QtCharts namespace incompatibility
The QtCharts.QtCharts namespace was removed in Qt6, and the contents are now directly in the QtCharts module.
To preserve compatibility with Qt5 code, the below works:
```
qtpy.QtCharts.QtCharts = qtpy.QtCharts
```
Maybe this is worth adding in qtpy itself?

dalthviz:
Fixes #304

Note: Needs a rebase after #313 is merged

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -211,9 +211,13 @@ try:
 except (ImportError, PythonQtError):
     pass
 
-# QtCharts backward compatibility with QtPy 1.x
-from . import QtCharts
-QtCharts.QtCharts = QtCharts
+try:
+    # QtCharts backward compatibility with QtPy 1.x
+    from . import QtCharts
+    QtCharts.QtCharts = QtCharts
+except (ImportError, PythonQtError):
+    pass
+
 
 def _warn_old_minor_version(name, old_version, min_version):
     """Warn if using a Qt or binding version no longer supported by QtPy."""
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -211,13 +211,6 @@ try:
 except (ImportError, PythonQtError):
     pass
 
-try:
-    # QtCharts backward compatibility with QtPy 1.x
-    from . import QtCharts
-    QtCharts.QtCharts = QtCharts
-except (ImportError, PythonQtError):
-    pass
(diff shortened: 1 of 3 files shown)


CAM-Gerlach (changes requested):
Because you're importing QtCharts in __init__, this makes the separate QtCharts module a hard dependency of QtPy, instead of just users importing `qtpy.QtCharts`, This also causes our tests to fail on platforms that don't have it by default, on which the test would normally be skipped (aside from the coverage-related failure you fixed in #313 ). Can you do this manipulation inside `qtpy.QtCharts` instead of in `__init__ to resolve this issue? Thanks!

CAM-Gerlach:
@dalthviz Just curious—is there a reason for using the the manual `[WiP]` title "tag" instead of marking as draft?

dalthviz:
> @dalthviz Just curious—is there a reason for using the the manual [WiP] title "tag" instead of marking as draft?

Pure habit basically

qtpy/__init__.py
```diff
@@ -211,6 +211,13 @@ class PythonQtWarning(Warning):
 except (ImportError, PythonQtError):
     pass
 
+try:
+    # QtCharts backward compatibility with QtPy 1.x
+    from . import QtCharts
+    QtCharts.QtCharts = QtCharts
+except (ImportError, PythonQtError):
+    pass
```
ccordoba12, line 219: I agree with @CAM-Gerlach: can this manipulation be done in the QtCharts module instead of here?

CAM-Gerlach: :point_up_2: ?

dalthviz: Not sure, I will need to check

ccordoba12 (approved):
Looks good to me now, thanks @dalthviz!

CAM-Gerlach (approved):
LGTM, thanks @dalthviz !

qtpy/__init__.py
CAM-Gerlach: Looks good!

PR #318, opened 2022-01-24
Patch QGuiApplication.desktop() for PySide6

nuno-andre:
`QGuiApplication.desktop()` is no longer available in PySide.

From PySide6's [_Class/Function Deprecations_][1]:

> `QDesktopWidget` has been removed. `QScreen` should be used instead, which can be retrieved using `QWidget.screen()`, `QGuiApplication.primaryScreen()` or` QGuiApplication.screens()`.

`QScreen` does not completely mimics `QDesktopWidget`, so further patching is still required.

[1]: https://doc.qt.io/qtforpython/porting_from2.html#class-function-deprecations

### DIFF
--- a/qtpy/QtGui.py
+++ b/qtpy/QtGui.py
@@ -33,7 +33,10 @@ elif PYSIDE2:
 elif PYSIDE6:
     from PySide6.QtGui import *
     from PySide6.QtOpenGL import *
+
+    # Map missing/renamed methods
     QFontMetrics.width = lambda self, *args, **kwargs: self.horizontalAdvance(*args, **kwargs)
+    QGuiApplication.desktop = lambda self: self.primaryScreen()
 
     # Map DeprecationWarning methods
     QDrag.exec_ = lambda self, *args, **kwargs: self.exec(*args, **kwargs)


CAM-Gerlach:
> `QScreen` does not completely mimics `QDesktopWidget`, so further patching is still required.

Hey @nuno-andre , thanks for letting us know! Do you plan to add this to your PR? I think it would be best ensure that the API is compatible (or at least nearly so) if we're going to patch it.

dalthviz:
Hi @nuno-andre thank you for the feedback on this and giving a try to an implementation! However as you say `QScreen does not completely mimics QDesktopWidget` so I'm not totally sure if we can do something robust enought over the QtPy side to provide complete compatibility with the `QDesktopWidget` definition of methods and signals . If we try to do it, probably we will need to create a `QDesktopWidget` class definition that maps methods/signals from `QGuiApplication`/`QScreen` and even with that I'm not totally sure that the functionality will be a one to one exact match with the way the removed class worked.

Taking this into account, and If you are working towards an application that supports Qt5 and 6, I would suggest to move from using `QDesktopWidget/QApplication.desktop()` to the different methods from `QGuiApplication` and `QScreen` (non deprecated and available for both versions).

Just in case, what do you think about this @ccordoba12 ?

CAM-Gerlach:
> Yeah, I would also like to see a full compatibility for QDesktopWidget in QtPy.

Just FYI @ccordoba12 , @dalthviz was suggesting that users move to `QScreen`, available in both Qt5 and Qt6, in lieu of QtPy backporting it to the legacy `QDesktopWidget`, i.e. the opposite of what you appear to be suggesting.

For what its worth, @dalthviz 's proposal sounds more reasonable to me, though I'd defer to whatever you two agree on.

ccordoba12:
What I meant is that if we were going to add support for `QDesktopWidget` here, we should support full compatibility with it.

nuno-andre:
@CAM-Gerlach @dalthviz @ccordoba12
This PR was only intended as a temporary solution, but I've just realized that it doesn't really solve any problems, since the only common method between Qt5's `QDesktopWidget` and Qt6's `QScreen` (`availableGeometry()`) doesn't even have the same signature. So I'm closing it.

Thank you all!

> @dalthviz [...] If you are working towards an application that supports Qt5 and 6, I would suggest to move from using QDesktopWidget/QApplication.desktop() to the different methods from QGuiApplication and QScreen (non deprecated and available for both versions).

Not my code. This is actually a bug I've found embedding [_qtconsole_](https://github.com/jupyter/qtconsole) in a PySide6 application.

CAM-Gerlach:
> Not my code. This is actually a bug I've found embedding qtconsole in a PySide6 application.

@nuno-andre So _our_ code, then :laughing: The Spyder team maintains QtConsole, so if you could report the issue there if you haven't already, that would be much appreciated. Thanks!

nuno-andre:
@CAM-Gerlach LOL, I didn't know. Sure, I'll open an issue.

Great work, btw ;)

CAM-Gerlach:
Just to be clear, I haven't contributed much of substance myself to the QtConsole code base, that is mostly @ccordoba12 , @dalthviz and others—so thank them, not me :)

PR #314, opened 2022-01-14
PR: Add mapping for missing enum values aliases on `QtCore.Qt`

dalthviz:
Fixes #306 

Note: Needs a rebase after #313 is merged

Issue #306, opened 2021-12-23
Qt6: missing unscoped enum values


thatfuckingbird:
Using PyQt6 via qtpy, QtCore.Qt.BackButton is not defined. The equivalent QtCore.Qt.XButton1 is defined, however (both are defined in PyQt5).
These are in the Qt::MouseButton enum, and both values are present in both Qt5 and Qt6 according to the Qt docs.

dalthviz:
Thank you again for the feedback @thatfuckingbird ! Indeed, there are some enum values that changed names that need adding an alias. Would you like to helps us with that? Let us know!

thatfuckingbird:
To be clear, these did *not* change name in Qt itself.

e.g. Qt::BackButton is listed on both of these pages:

https://doc.qt.io/qt-6/qt.html
https://doc.qt.io/qt-5/qt.html

Qt::BackButton and Qt::ForwardButton are the only 2 I encountered, but there could be others. Now I suspect they are missing because Python Enums don't support duplicate values (?). In that case sadly I don't see how we could fix it.

dalthviz:
At least for the unscoped access I think we could do some aliasing after promoting enums like this:

In `qtpy/QtCore.py` after the lines

```python
from .enums_compat import promote_enums
promote_enums(QtCore)
del QtCore
```

do

```python
Qt.BackButton = Qt.XButton1
Qt.ForwardButton = Qt.XButton2
```

CAM-Gerlach:
> Now I suspect they are missing because Python Enums don't support duplicate values (?)

Only if you use the `@enum.unique` decorator on them. This should be fixable by upstream if so; did you try reporting it there?

PR #319, opened 2022-01-28
PR: Promote enum aliases

MatthieuDartiailh:
A number of enum values in QT can be accessed through aliases. For example, QSlider.TicksPosition has both TicksAbove  and TicksLeft and TicksLeft is an alias of TicksAbove.

Currently, enums are only iterated which yield their variants but miss the existence of aliases. This patch explicitly iterate over the enum members to catch all aliases. Tests are included, both for TicksLeft which is a documented alias (https://doc.qt.io/qt-6/qslider.html#TickPosition-enum) and SC_SliderGroove which happens to have the same value as SC_ScrollBarAddLine and hence appears as an alias.

### DIFF
--- a/qtpy/QtCore.py
+++ b/qtpy/QtCore.py
@@ -48,10 +48,6 @@ if PYQT6:
     promote_enums(QtCore)
     del QtCore
 
-    # Map missing unscoped access enum values
-    Qt.BackButton = Qt.XButton1
-    Qt.ForwardButton = Qt.XButton2
-
     # Alias deprecated ItemDataRole enum values removed in Qt6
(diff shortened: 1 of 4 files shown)


dalthviz (commented):
Thank you @MatthieuDartiailh for helping with this!

Could you remove also in this PR the manual aliases done here?:  https://github.com/spyder-ide/qtpy/blob/74dd30998e01b6058af41d2237af971d68b46cae/qtpy/QtCore.py#L51-L53 

With this change I think those lines are now unnecessary.

Note: Some related issues/PRs: #306 #314

MatthieuDartiailh:
Done

dalthviz (approved):
Thank you @MatthieuDartiailh ! :+1:

PR #321, opened 2022-02-07
Add missing alias for Qt.MidButton for pyqt6

Tillsten:
See topic


Tillsten:
Sadly this does not work

PR #324, opened 2022-02-26
PR: Restrict broken Pytest versions to those not affected by the Pytest 7.0.0 import-mode=importlib behavior regression

CAM-Gerlach:
Presently, Pytest's ``--importmode=importlib`` causes our test suite to fail at collection due to a regression introduced in Pytest 7.0.0, pytest-dev/pytest#9645 . This was fixed in pytest-dev/pytest#9681 , but hasn't yet made it into a Pytest release. Therefore, we work around it for now by disabling ``--importmode=importlib``.

Fixes #324


dalthviz (approved):

PR #334, opened 2022-04-08
PR: Remove `QHeaderView` patch related files

### LINKED ISSUE #332: Remove `QHeaderView` patch
Checking a little bit I think the current logic for the `QHeaderView` patch was done in order to give support to removed methods from PyQt4/PySide/Qt4 (https://github.com/spyder-ide/qtpy/issues/64 https://github.com/spyder-ide/qtpy/pull/65) and rigth now the test for the patch is being skipped in a good amount of testing setups since segfaults with Qt 5.15+  Qt6 and also with Python 3.7+ on MacOS

Taking this into account, maybe we should consider removing this patch logic? What do you think @ccordoba12 @CAM-Gerlach ?

dalthviz:
Fixes #332


ccordoba12 (approved):

dalthviz:
Do you have any comment @CAM-Gerlach ?

CAM-Gerlach (approved):
LGTM, sorry for the delay @dalthviz —I was taking a short break from GitHub for a few days to focus on other things, and then got overwhelmed when I came back

CAM-Gerlach:
Looks like the tests started segfaulting on PyQt6 + Windows on both Py3.7 and 3.10 (no conda). Based on the check logs, it seems likely this is due to the PyQt PyPI wheels for Qt 6.3.0 being released; any idea on fixing this from our end, or should we pin to qt 6.2.x for now?

dalthviz:
> any idea on fixing this from our end, or should we pin to qt 6.2.x for now?

Maybe it make sense to open a new issue for that and check if we can fix it for 2.1.0 (hopefully preventing the segfault even when using Qt 6.3.0 instead of just constraining to install Qt 6.2.x) Will open a new issue with your comment and start checking if I can do something about it :+1:

CAM-Gerlach:
Thanks @dalthviz !

PR #335, opened 2022-04-08
PR: Fix uic skipped tests and PyQt 6.3.0 segfaulting tests

dalthviz:
Fixes #265 
Fixes #336

PR #337, opened 2022-04-19
PR: Add command line support for Mypy

### LINKED ISSUE #216: Support type hints and mypy

CAM-Gerlach:
Adds a CLI to inject the appropriate arguments into a user's Mypy invocation to ensure proper type checking. Also provides an easy way to check the currently selected Qt binding and QtPy's version.

This is based on #217 , rebased and updated for QtPy 2.x, with a number of other improvements, fixes and refinements.

Fixes #216 
Closes #217


CAM-Gerlach:
Going to need #335 for tests to pass here, BTW

qtpy/__main__.py
```diff
@@ -0,0 +1,9 @@
+import qtpy.cli
```
dalthviz, line 1: Should we add a copyright/license header here?

README.md
```diff
@@ -71,6 +71,25 @@ conda install qtpy
 ```
 
 
+### Mypy
+
+A CLI is offered to help with usage of QtPy.
+Presently, the only feature is to generate command line arguments for Mypy
+that will enable it to process the QtPy source files with the same API
+as QtPy itself would have selected.
```
dalthviz, line 79: Maybe add here that the example result below is the one you get when running QtPy with PySide2 as the available/selected binding.

dalthviz: Also, checking this inside an env without bindings installed probably we should also add somewhere that the default output for this command is `--always-true=PYQT5 --always-false=PYQT6 --always-false=PYSIDE2 --always-false=PYSIDE6` ?

dalthviz (commented):
Thanks @CAM-Gerlach ! Left some comments regarding the description in the README for the command and a missing header for a file otherwise this LGTM :+1:

qtpy/__main__.py
CAM-Gerlach: We could, though I didn't because the content is just a snippit of standard boilerplate that is neither substantive nor original enough to actually be copyrightable in the first case. In any case, I added it along with a module-level docstring, and also fixed the no-longer-needed in Python 3 encoding line added to the other file.

README.md
CAM-Gerlach: Good point, the existing description here was not very good, and rather incomplete and potentially confusing. I expanded it to explicitly mention both those things as well as the actual example CLI invocation and other clarifications.

dalthviz (approved):
:+1:

ccordoba12 (commented):
Thanks @CAM-Gerlach for your help with this! I left some minor comments, otherwise looks good to me.

README.md
```diff
@@ -71,6 +71,37 @@ conda install qtpy
...
+
+If you run
+
+```bash
+qtpy mypy-args
+```
+
+QtPy will output a string of Mypy CLI args that will reflect the currently
+selected Qt API.
+For example, in an environment where `PYQT5` would be selected
```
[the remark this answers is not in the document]
CAM-Gerlach: If so, then it shouldn't be in `code`, so:

```suggestion
For example, in an environment where PyQt5 would be selected
```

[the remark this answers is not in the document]
ccordoba12: Ok, no problem with that.

ccordoba12 (approved):
Thanks @CAM-Gerlach!

Issue #154, opened 2018-05-21
Circular attempt for bindings


dpizetta:
Hi, I'd like to implement a change for setup the binding in a circular form. Here is an example.

If the preference list is [pyqt5, pyside2, pyqt, pyside]:
- choosing pyqt, if it is not found, it should try pyside, pyqt5, pyside2;
- choosing pyside2, if it is not found, it should try pyqt, pyside2, pyqt5;

I think this would improve the auto setup and even the message that the "qtpy.PythonQtError: No Qt bindings could be found", because if you choose the last one, all the previous are not tested, so it is not "not found" actually.

Let me know what do you think about it, tks

goanpeca:
@dpizetta it sounds like a good idea :-) @ccordoba12 ?

goanpeca:
I would add that we need to add an else check so that if the last item in the list is not found, it should say something regarding that, instead of saying that it failed to use whatever binding check came last. Pyside at the moment, which is misleading

ccordoba12:
@dpizetta, I like it! But things should not be circular. Instead we should always prefer the bindings for the same Qt major version, then the other ones. And the PyQt bindings should be preferred over the PySide ones because they are more stable.

So, according to what you post above:

* choosing pyside, if it is not found, it should try pyqt, pyqt5, pyside2;
* choosing pyqt, if it is not found, it should try pyside, pyqt5, pyside2;
* choosing pyside2, if it is not found, it should try pyqt5, pyqt, pyside

dpizetta:
Hi @goanpeca , @ccordoba12 thank you. I agree with your comments. Implementing this proposal also include the additional feature from @goanpeca. I'll post PR soon.

goanpeca:
Thanks @dpizetta :-) !

dpizetta:
@goanpeca  > **Just for context, is there any particular issues on your side that would directly benefit from this? Like an example of what was the problem?**
Sorry, I miss this question. The main problem is: if you choose the last one, i.e PySide, but it is not installed, it wouldn't try the other ones. So I think it should be nice to have this feature. In addition, I would like to add some functions that can be useful for other users, ex. get_available_api, that returns installed qt bindings..etc. I practically finished the code, but I would like to add more information about the changes to be well documented.

ccordoba12:
I think what he meant is that if all four bindings are tried and none is found, then `qtpy` should print a message saying precisely that.

[...]

PR #340, opened 2022-04-26
PR: Fix bug with environ handling

larsoner:
I'm not 100% sure what the intention of modifying `os.environ` at all is, but assuming it's meant to set it so that e.g. new processes that spawn make use of the same framework, then there's a bug in the current implementation in `master`:
```
>>> import os
>>> import PyQt6
>>> 'QT_API' in os.environ
False
>>> import qtpy
>>> qtpy.API  # good
'pyqt6'
>>> 'QT_API' in os.environ  # okay...
True
>>> os.environ['QT_API']  # bad
'pyqt5'
```
This bug was found precisely because a pattern like this: have no QT_API set, import PyQt6, then import qtpy, then spawn a new process (using joblib), have it fail because in that process `import qtpy` fails because PyQt5 isn't installed, but `os.environ['QT_API'] == 'pyqt5'`.

This PR thus avoids changing `os.environ` at all until a suitable API has been chosen, and only then (in the `else` clause of the `try`s) set `os.environ[QT_API]`.

An alternative would be not to modify `os.environ` at all, but this seems less than ideal because it will make it so that the spawned process will not necessarily end up with the same Qt binding, depending on whether or not that process does an `import PyQt6` and `import qtpy` in the same order as the main process, at least in the case where a user has two bindings (e.g., PyQt5 and PyQt6).

This PR also promotes a `assert ` statement to a proper and more informative `raise ValueError(...)` (that also will not be skipped in Python optimized mode), and changes some `'QT_API'`s to `QT_API`, as it seems like that's what the variable name is for. But if either of these is not actually desirable, I can revert!

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -72,6 +72,10 @@ class PythonQtWarning(Warning):
     """Warning if some features are not implemented in a binding."""
 
 
+class PythonQtValueError(ValueError):
+    """Error raised if an invalid QT_API is specified."""
+
+
 # Qt API environment variable name
 QT_API = 'QT_API'
 
@@ -99,12 +103,13 @@ PYSIDE_VERSION_MIN = PYSIDE2_VERSION_MIN
 binding_specified = QT_API in os.environ
 
 API_NAMES = {'pyqt5': 'PyQt5', 'pyqt6': 'PyQt6',
-             'pyside2':'PySide2', 'pyside6': 'PySide6'}
+             'pyside2': 'PySide2', 'pyside6': 'PySide6'}
 API = os.environ.get(QT_API, 'pyqt5').lower()
 initial_api = API
 if API not in API_NAMES:
-    raise ValueError(
-        f'Specified QT_API={repr(QT_API)} not in valid options: {API_NAMES}')
+    raise PythonQtValueError(
+        f'Specified QT_API={repr(QT_API.lower())} not in valid options: '
+        f'{API_NAMES}')
 
 is_old_pyqt = is_pyqt46 = False
 QT5 = PYQT5 = True
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -98,14 +98,13 @@ PYSIDE_VERSION_MIN = PYSIDE2_VERSION_MIN
 # Detecting if a binding was specified by the user
 binding_specified = QT_API in os.environ
 
-# Setting a default value for QT_API
-os.environ.setdefault(QT_API, 'pyqt5')
-
 API_NAMES = {'pyqt5': 'PyQt5', 'pyqt6': 'PyQt6',
              'pyside2':'PySide2', 'pyside6': 'PySide6'}
-API = os.environ[QT_API].lower()
+API = os.environ.get(QT_API, 'pyqt5').lower()
 initial_api = API
-assert API in API_NAMES
+if API not in API_NAMES:
+    raise ValueError(
+        f'Specified QT_API={repr(QT_API)} not in valid options: {API_NAMES}')
 
 is_old_pyqt = is_pyqt46 = False
 QT5 = PYQT5 = True
@@ -150,7 +149,9 @@ if API in PYQT5_API:
 
             del macos_version
     except ImportError:
-        API = os.environ['QT_API'] = 'pyqt6'
+        API = 'pyqt6'
+    else:
+        os.environ['QT_API'] = API
 
 if API in PYQT6_API:
     try:
@@ -161,7 +162,9 @@ if API in PYQT6_API:
         QT6 = PYQT6 = True
 
     except ImportError:
-        API = os.environ['QT_API'] = 'pyside2'
+        API = 'pyside2'
+    else:
+        os.environ['QT_API'] = API
 
 
 if API in PYSIDE2_API:
@@ -183,7 +186,9 @@ if API in PYSIDE2_API:
 
             del macos_version
     except ImportError:
-        API = os.environ['QT_API'] = 'pyside6'
+        API = 'pyside6'
+    else:
+        os.environ['QT_API'] = API
 
 if API in PYSIDE6_API:
     try:
@@ -194,7 +199,9 @@ if API in PYSIDE6_API:
         QT6 = PYSIDE6 = True
 
     except ImportError:
-        API = os.environ['QT_API'] = 'pyqt5'
+        API = 'pyqt5'
+    else:
+        os.environ['QT_API'] = API
 
 
 # If a correct API name is passed to QT_API and it could not be found,
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -156,7 +156,7 @@ if API in PYQT5_API:
     except ImportError:
         API = 'pyqt6'
     else:
-        os.environ['QT_API'] = API
+        os.environ[QT_API] = API
 
 if API in PYQT6_API:
     try:
@@ -169,7 +169,7 @@ if API in PYQT6_API:
     except ImportError:
         API = 'pyside2'
     else:
-        os.environ['QT_API'] = API
+        os.environ[QT_API] = API
 
 
 if API in PYSIDE2_API:
@@ -193,7 +193,7 @@ if API in PYSIDE2_API:
     except ImportError:
         API = 'pyside6'
     else:
-        os.environ['QT_API'] = API
+        os.environ[QT_API] = API
 
 if API in PYSIDE6_API:
     try:
@@ -206,7 +206,7 @@ if API in PYSIDE6_API:
     except ImportError:
         API = 'pyqt5'
     else:
-        os.environ['QT_API'] = API
+        os.environ[QT_API] = API
 
 
 # If a correct API name is passed to QT_API and it could not be found,
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -108,7 +108,7 @@ API = os.environ.get(QT_API, 'pyqt5').lower()
 initial_api = API
 if API not in API_NAMES:
     raise PythonQtValueError(
-        f'Specified QT_API={repr(QT_API.lower())} not in valid options: '
+        f'Specified QT_API={repr(QT_API.lower())} is not in valid options: '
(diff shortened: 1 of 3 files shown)


qtpy/__init__.py
```diff
@@ -98,14 +98,13 @@ class PythonQtWarning(Warning):
...
-os.environ.setdefault(QT_API, 'pyqt5')
-
 API_NAMES = {'pyqt5': 'PyQt5', 'pyqt6': 'PyQt6',
              'pyside2':'PySide2', 'pyside6': 'PySide6'}
-API = os.environ[QT_API].lower()
+API = os.environ.get(QT_API, 'pyqt5').lower()
 initial_api = API
-assert API in API_NAMES
+if API not in API_NAMES:
+    raise ValueError(
```
dalthviz, line 106: Maybe we could use here a custom exception like `PythonQtValueError(ValueError)`?

dalthviz (commented):
Thank you @larsoner for finding the bug and checking into this! I left a comment regarding the `ValueError` exception (maybe we shoudl create a new custom exception following the spirit of `PythonQtError` and `PythonQtWarning` ).

Also, could be worthy to try to add a test to check that we now have the proper behavior (the env var `QT_API` is updated correctly when is not initially set). I think the test should go in [`qtpy/tests/test_main.py`](https://github.com/spyder-ide/qtpy/blob/master/qtpy/tests/test_main.py).

Besides that this LGTM :+1:

qtpy/tests/conftest.py
```diff
@@ -21,8 +21,9 @@ def pytest_report_header(config):
     versions += 'PyQt6: '
 
     try:
-        from PyQt6 import Qt
-        versions += f"PyQt: {Qt.PYQT_VERSION_STR} - Qt: {Qt.QT_VERSION_STR}"
+        from PyQt6 import QtCore
+        versions += \
+            f"PyQt: {QtCore.PYQT_VERSION_STR} - Qt: {QtCore.QT_VERSION_STR}"
```
larsoner, line 26: I needed this on my system at least to get PyQt6 to actually show up

qtpy/tests/test_main.py
```diff
@@ -16,6 +20,7 @@ def assert_pyside2():
     assert QtGui.QPainter is PySide2.QtGui.QPainter
     assert QtWidgets.QWidget is PySide2.QtWidgets.QWidget
     assert QtWebEngineWidgets.QWebEnginePage is PySide2.QtWebEngineWidgets.QWebEnginePage
+    assert os.environ['QT_API'] == 'PySide2'
```
dalthviz, line 23: I think the tests are failing due to lower vs uppercase names for the bindings
```suggestion
    assert os.environ['QT_API'] == 'pyside2'
```

qtpy/tests/test_main.py
```diff
@@ -27,6 +32,7 @@ def assert_pyside6():
     assert QtWidgets.QWidget is PySide6.QtWidgets.QWidget
     # Only valid for qt>=6.2
     # assert QtWebEngineWidgets.QWebEnginePage is PySide6.QtWebEngineCore.QWebEnginePage
+    assert os.environ['QT_API'] == 'PySide6'
```
dalthviz, line 35:
```suggestion
    assert os.environ['QT_API'] == 'pyside6'
```

qtpy/tests/test_main.py
```diff
@@ -40,6 +46,7 @@ def assert_pyqt5():
         assert QtWebEngineWidgets.QWebEnginePage is PyQt5.QtWebEngineWidgets.QWebEnginePage
     else:
         assert QtWebEngineWidgets.QWebEnginePage is PyQt5.QtWebKitWidgets.QWebPage
+    assert os.environ['QT_API'] == 'PyQt5'
 
 def assert_pyqt6():
```
dalthviz, line 51: I guess something like `assert os.environ['QT_API'] == 'pyqt6'` is missing for the `assert_pyqt6` function

larsoner:
Green!

qtpy/__init__.py
```diff
@@ -161,7 +167,9 @@ class PythonQtWarning(Warning):
         QT6 = PYQT6 = True
 
     except ImportError:
-        API = os.environ['QT_API'] = 'pyside2'
+        API = 'pyside2'
```
ccordoba12, line 170: In the same line with the above, this should `pyside6` for me.

qtpy/__init__.py
```diff
@@ -183,7 +191,9 @@ class PythonQtWarning(Warning):
 
             del macos_version
     except ImportError:
-        API = os.environ['QT_API'] = 'pyside6'
+        API = 'pyside6'
```
ccordoba12, line 194: This should be `pyqt5`.

qtpy/__init__.py
```diff
@@ -194,7 +204,9 @@ class PythonQtWarning(Warning):
         QT6 = PYSIDE6 = True
 
     except ImportError:
-        API = os.environ['QT_API'] = 'pyqt5'
+        API = 'pyqt5'
```
ccordoba12, line 207: And this `pyqt6`.

qtpy/tests/test_main.py
```diff
@@ -83,3 +91,39 @@ def test_qt_api():
                 assert_pyqt6()
         else:
             assert_pyqt5()
+
+
+@pytest.mark.parametrize('api', API_NAMES.values())
+def test_qt_api_environ(api):
+    """
+    If no QT_API is specified but some Qt is imported, ensure environ is set properly
```
ccordoba12, line 99:
```suggestion
    If no QT_API is specified but some Qt is imported, ensure QT_API is set properly.
```

qtpy/tests/test_main.py
```diff
@@ -1,6 +1,10 @@
 import os
+import sys
+import subprocess
 
-from qtpy import QtCore, QtGui, QtWidgets
+import pytest
+
+from qtpy import QtCore, QtGui, QtWidgets, API_NAMES, PythonQtValueError
 try:
```
ccordoba12, line 8:
```suggestion

try:
```

Adding a blank line before this one for clarity.

ccordoba12 (commented):
Thanks for your work on this @larsoner! I left some small comments for you below.

qtpy/__init__.py
```diff
@@ -150,7 +154,9 @@ class PythonQtWarning(Warning):
 
             del macos_version
     except ImportError:
-        API = os.environ['QT_API'] = 'pyqt6'
+        API = 'pyqt6'
```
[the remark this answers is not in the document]
larsoner: Not sure, I just tried not to change it in this PR.

From a quick look, the default is PyQt5 (i.e., it's what API is if QT_API is not set), and there currently is no differentiation between whether a user had set it or not at this point in the code. And the docs say the bindings will be tried in some specific order, and I think that order is meant to be followed in the code.

Given that this stuff is not trivial, and the fallback order is (or at least should be!) unchanged in this PR, I'd rather not tackle that here, and leave that for another PR, probably for someone else after some additional discussion. For example, if I just committed all the suggested changes about which API should come next, I don't think the code will work as desired -- it's currently framed around an API "try order" that gets followed, so below when you suggest changing the pyside2-fail-path (line 194) to set the `API = 'pyqt5'` instead of `API = 'pyside6'`, this will not have the intended (I think?) effect of actually *trying* pyqt5, because that happens earlier in the code (up here starting around line 130), whereas `'pyside6'` does currently get tried because it comes later in the code path... so I think more time/thought/effort/refactoring would be necessary to achieve the goal of the suggested changes.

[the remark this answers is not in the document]
dalthviz: I think the change @ccordoba12 wanted to suggest is changing the _`try order`_ to be `PyQt5, PySide2, PyQt6, PySide6` (which will need not only to change the values on the `else` statements but also moving the whole blocks for the respective binding to follow such order). However, as @larsoner said I think that's kind of out the scope for this PR. The full behavior described (some sort of intelligent fallback) maybe its related with #154 (so if you try PyQt5 fallback to PySide2, but also if you try PySide2 fallback to PyQt5 and so on).

Just in case @ccordoba12 , I think the _`try order`_ on `v1.x` was something like `PyQt5, PySide2, PyQt4, PySide` but after the removal of Qt4 and addition of Qt6 the order on `v2.x/master` is now `PyQt5, PyQt6, PySide2, PySide6`. So maybe opening a new issue for that could be worthy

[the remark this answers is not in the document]
ccordoba12: If the current order is `PyQt5, PyQt6, PySide2, PySide6`, then I agree it shouldn't be changed on this PR.

Sorry for the noise guys, ignore my comments.

qtpy/tests/conftest.py
dalthviz: Regarding this change, you didn't get shown the version info when testing locally? Could you try to restore this line and check if the issue is still presented (not seeing the version header in the pytest report) locally? Thanks!

ccordoba12 (approved):
Thanks @larsoner for your help with this!

qtpy/tests/conftest.py
larsoner: > Regarding this change, you didn't get shown the version info when testing locally? Could you try to restore this line and check if the issue is still presented (not seeing the version header in the pytest report) locally? Thanks!

If I revert this then I won't get the version:
```
$ python -c "from PyQt6 import Qt"
  File "<string>", line 1, in <module>
ImportError: cannot import name 'Qt' from 'PyQt6' (/home/larsoner/.local/lib/python3.10/site-packages/PyQt6/__init__.py)
```

dalthviz (approved):
Thank you for all the work here @larsoner !

larsoner:
Happy to contribute to this awesome package! We've successfully used it to seamlessly transition three packages from PyQt5 only to framework-agnostic code

larsoner:
- [MNE-Python](https://github.com/mne-tools/mne-python/pull/10430)
- [PyVistaQt](https://github.com/pyvista/pyvistaqt/pull/61)
- [mne-qt-browser](https://github.com/mne-tools/mne-qt-browser/pull/104)

PR #341, opened 2022-04-28
PR: Add workaround for `mode` argument in QTextCursor.movePosition (Pyside2)

rear1019:
See commit message/comment in code for details.

PR #346, opened 2022-05-20
PR: Add workaround for `mode` argument in QTextCursor.movePosition (PySide6)

rear1019:
Also see PR #341


dalthviz (approved):
Thanks @rear1019 for your help with this!

PR #344, opened 2022-05-01
PR: Add missing imports and modules

DaelonSuzuka:
QtWebSockets was not being imported when using PySide6.

Some other modules appear to missing imports as well:

- Qt3DAnimation.py
- QtDBus.py
- QtDesigner.py
- QtLocation.py
- QtNetworkAuth.py
- QtOpenGlWidgets.py
- QtPositioning.py
- QtTextToSpeech.py
- QtWebEngine.py
- QtXmlPatterns.py

It's possible that some of these were omitted on purpose, but there's no indication of that in the code.

It would also help readability and auditability to standardize the formatting of these files. I'd be willing to do this if the project is open to it.

Additionally, PySide6 has been split into two packages, PySide6-Essentials and PySide6-Addons, and it's probably incorrect behavior to assume that all of PySide6 is present without explicitly checking.

PR #350, opened 2022-06-13
PR: Restore `WEBENGINE` constant in `QtWebEngineWidgets`

ccordoba12:
This constant was removed in PR #344, but it's used by external projects (e.g. Spyder). So, they would break if they can find it.


dalthviz (approved):
Thanks @ccordoba12 LGTM :+1:

PR #353, opened 2022-06-30
PR: Add note to readme about use with Pyright

### LINKED ISSUE #352: FR: Deprecation Warning for Enum Access
Per [this comment here](https://github.com/napari/napari/issues/3263#issuecomment-907159978):

> starting with sip v4.19.9 (around PyQt 5.11) all enums are now scoped, and starting in PyQt6, the backwards compatibility allowing enums to be accessed in their top level namespace has been removed

This is confirmed by the [PyQt5 Gotcha's docs](https://www.riverbankcomputing.com/static/Docs/PyQt5/gotchas.html#enums).

Since end users may be accessing different versions of PyQt/PySide through `qtpy`, would it be possible to add a `DeprecationWarning` for the corresponding versions? (most likely this could be determined at runtime via `QT_API`)

e.g. I noticed a `"Type[QEvent]" has no attribute "Gesture"` mypy error in legacy code where `Gesture` was accessed like so:

image

and the fix was instead to use `QtCore.QEvent.Type.Gesture`.

Perhaps something like:

```
DeprecationWarning: Enums are scoped beginning with PyQt 5.11 (and its corresponding PySide 2 version). Access to enums through top-level namespaces has been deprecated beginning in PyQt 6.

Use the full path to the enumeration instead. For example:

QtCore.QEvent.Type.Gesture

instead of 

QtCore.QEvent.Gesture
```

CAM-Gerlach:
Per discussion in #352 , this adds a note to the readme about usage with Pyright under the Type Checker Integration section. Thanks, @adam-grant-hendry!

Resolves #352

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ conda install qtpy
 ```
 
 
-### Mypy integration
+### Type checker integration
 
 A Command Line Interface (CLI) is offered to help with usage of QtPy.
 Presently, its only feature is to generate command line arguments for Mypy
@@ -100,6 +100,8 @@ the Mypy command line invocation as follows:
 ```bash
 mypy --package mypackage $(qtpy mypy-args)
 ```
+
+For Pyright support and other usage notes, see [this comment](https://github.com/spyder-ide/qtpy/issues/352#issuecomment-1170684412).
 
 
 ## Contributing


dalthviz (approved):
Thanks @CAM-Gerlach and @adam-grant-hendry !

PR #355, opened 2022-07-16
compat.py: Add utility function to get raw bytes from QImages

zjp:
We use this utility function in ChimeraX to grab the raw content of a QImage to pass to our renderer. I'm hoping it will be useful to other users as well, and I'm open to feedback about how to make that possible.

### DIFF
--- a/qtpy/compat.py
+++ b/qtpy/compat.py
@@ -129,3 +129,15 @@ def getsavefilename(parent=None, caption='', basedir='', filters='',
                                 caption=caption, basedir=basedir,
                                 filters=filters, selectedfilter=selectedfilter,
                                 options=options)
+
+# =============================================================================
+def imagebytes(qimage):
+    """Get the content of a QImage as raw bytes for use in e.g. rendering."""
+    if PYQT5:
+        return qimage.bits().asstring(qimage.byteCount())
+    elif PYQT6:
+        return qimage.bits().asstring(qimage.sizeInBytes())
+    elif PYSIDE2 or PYSIDE6:
+        return qimage.bits().tobytes()
+    else:
+        raise QtBindingsNotFoundError()


dalthviz:
Thanks @zjp ! Then this is comming from ChimeraX source code, right? Is there any licensing we should know about? 

If that is the case I would say that the license should be compatible with MIT (the license we use here) and we should add a note in the function pointing out from where the function is comming from.

Besides that adding a tests for this could be nice too :)

Maybe @CAM-Gerlach and/or @ccordoba12 have more elements to have in mind.

zjp:
There's an equivalent function but it needed adaptation to your project. There's no licensing concerns on your end -- my goal is to upstream this code and remove it from ChimeraX.

I'll add tests!

ccordoba12:
@zjp, this has a conflict with master now.

zjp:
When I get a chance I'll rebase thanks for letting me know!

qtpy/compat.py
```diff
@@ -129,3 +129,15 @@ def getsavefilename(parent=None, caption='', basedir='', filters='',
                                 caption=caption, basedir=basedir,
                                 filters=filters, selectedfilter=selectedfilter,
                                 options=options)
+
+# =============================================================================
+def imagebytes(qimage):
```
CAM-Gerlach, line 134:
```suggestion
def getimagebytes(qimage):
```

Unless there's a particular reason for this exact name, what about naming this function as a verb, per standard Python/PEP 8 convention as well as that used within this module?

CAM-Gerlach:
> Thanks @zjp ! Then this is comming from ChimeraX source code, right? Is there any licensing we should know about?

If @zjp is the author or otherwise holds the copyright, then by submitting it here they are agreeing to license it under QtPy's license, presuming that corporate contracts don't get in the way. Otherwise, there could theoretically be an issue, but while IANAL and this is not legal advice, the code is _probably_ too short/simple to be covered by copyright protection as an independent work.

> I'll add tests!

Looking forward to it!

qtpy/compat.py
zjp: Oversight on my part, I've made this change and the function is now called 'getimagebytes'.

dalthviz:
Thanks @zjp for giving this an update! Just in case, seems like the failing checks are related with https://github.com/coverallsapp/github-action/issues/205

zjp:
I still need to add a test for this function. Do you think it's better to add a small image to the repository, to fetch an image from the internet in the test, or to do a secret third thing?

dalthviz:
> Do you think it's better to add a small image to the repository, to fetch an image from the internet in the test, or to do a secret third thing?

Maybe we could create a `QImage` manually? Like a black square, maybe even save it and then use that to validate the function? So creating an image with something like:

```python
from qtpy.QtCore import QRectF, QSize, Qt
from qtpy.QtGui import QBrush, QImage, QPainter

image = QImage(QSize(100, 100), QImage.Format_RGB32)
painter = QPainter(image)
painter.setBrush(QBrush(Qt.black))
painter.fillRect(QRectF(0, 0, 100, 100), Qt.black)
image.save("black.png")
``` 

What do you think ? Also pinging @ccordoba12 just in case for more ideas

ccordoba12:
> What do you think ? Also pinging @ccordoba12 just in case for more ideas

I like it, I think it's a good idea.

zjp:
Sounds good to me.

zjp:
I've added the test. It seemed sensible to me to test the length of the byte array, but if you immediately think of a better condition let me know.

zjp:
I was trying to fill in the color of the QImage in the test the wrong way, which led to a segfault (though curiously not on a couple platforms). Anyway, I think the QImage doesn't need to be filled in at all. A byte representing color will be the same length whether it's (1,1,1,1) or (0,0,0,0).

PR #358, opened 2022-07-23
PR: Fix PyQt6 typing import for Qt

tlambert03:
small fix that allows IDEs to find QtCore.Qt in PyQt6

### DIFF
--- a/qtpy/QtCore.py
+++ b/qtpy/QtCore.py
@@ -8,6 +8,7 @@
 
 """Provides QtCore classes and functions."""
 from typing import TYPE_CHECKING
+
 from . import PYQT6, PYQT5, PYSIDE2, PYSIDE6, QtBindingsNotFoundError
 
 if PYQT5:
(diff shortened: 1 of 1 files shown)


qtpy/QtCore.py
```diff
@@ -7,7 +7,7 @@
 # -----------------------------------------------------------------------------
 
 """Provides QtCore classes and functions."""
-
+from typing import TYPE_CHECKING
 from . import PYQT6, PYQT5, PYSIDE2, PYSIDE6, QtBindingsNotFoundError
```
CAM-Gerlach, line 11:
```suggestion
from typing import TYPE_CHECKING

from . import PYQT6, PYQT5, PYSIDE2, PYSIDE6, QtBindingsNotFoundError
```

Should be in a separate import group (stdlib vs third party), per PEP 8 and standard convention.

CAM-Gerlach (commented):
One small style comment, otherwise LGTM at least.

qtpy/QtCore.py
tlambert03: sure!  have you considered pre-commit here?

CAM-Gerlach (approved):
LGTM, thanks @tlambert03 . @dalthviz ?

dalthviz (approved):
Thanks @tlambert03 !

PR #360, opened 2022-08-10
CI: Ensure conda-forge channel usage when testing with conda

dalthviz:
Seems like the env for Windows with Python 3.7 and PyQt 5.9 on the CI is not working

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -82,6 +82,7 @@ jobs:
            auto-activate-base: true
            auto-update-conda: true
            channels: conda-forge
+           channel-priority: strict
       - name: Print Conda info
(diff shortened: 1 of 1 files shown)


CAM-Gerlach (changes requested):
``channels: conda-forge`` is already set in the `setup-miniconda` step in the workflow, so the `-c conda-forge` is redundant here and has no additional effect (thus the test still fails).

Instead, if you drop this commit and just add `channel-priority: strict` to the aforementioned step, everything _should_ work as expected.

CAM-Gerlach (approved):
LGTM, thanks @dalthviz !

Since this is a small change and fixes the CIs, I'll go ahead and merge.

PR #361, opened 2022-08-11
PR: toPython helpers for QDate and QTime

bob-schumaker:
This adds toPython() compat functions to QDate and QTime for symmetry.

### DIFF
--- a/qtpy/QtCore.py
+++ b/qtpy/QtCore.py
@@ -21,7 +21,9 @@ if PYQT5:
 
     # For issue #153 and updated for issue #305
     from PyQt5.QtCore import QDateTime
+    QDate.toPython = lambda self, *args, **kwargs: self.toPyDate(*args, **kwargs)
     QDateTime.toPython = lambda self, *args, **kwargs: self.toPyDateTime(*args, **kwargs)
+    QTime.toPython = lambda self, *args, **kwargs: self.toPyTime(*args, **kwargs)
 
     # Map missing methods on PyQt5 5.12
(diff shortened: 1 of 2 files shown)


qtpy/QtCore.py
```diff
@@ -21,7 +21,9 @@
 
     # For issue #153 and updated for issue #305
     from PyQt5.QtCore import QDateTime
```
CAM-Gerlach, line 23:
```suggestion
    from PyQt5.QtCore import QDate, QDateTime, QTime
```

Shouldn't they be explicitly imported here too?

CAM-Gerlach (commented):
Seems naively reasonable to me overall, but I'd defer to @dalthviz on that.

Could you either add tests like the [one for `QDateTime.toPython`](https://github.com/spyder-ide/qtpy/blob/e21b92ff9a645d05311f7512aeb14c1fdd7f5697/qtpy/tests/test_qtcore.py#L25) (or parameterize that test accordingly)?

dalthviz (commented):
Hi @bob-schumaker thank you for working on this! I would say that the missing thing here is to add some tests for `QDate.toPython` and `QTime.toPython` as @CAM-Gerlach mentioned. If you need help with that let us know!

bob-schumaker:
> Hi @bob-schumaker thank you for working on this! I would say that the missing thing here is to add some tests for `QDate.toPython` and `QTime.toPython` as @CAM-Gerlach mentioned. If you need help with that let us know!
> 
> Other than that this LGTM 👍

Duh, out of practice. Let me add the tests :-)

dalthviz (approved):
Thanks @bob-schumaker !

PR #362, opened 2022-08-19
Fix typos

luzpaz:
Found via `codespell -q 3`


CAM-Gerlach (approved):
LGTM, thanks.

To note, we use `codespell` in the `pre-commit` config on several of our other repos to catch and fix typos like these automatically, which I am in the process of adding here, so these will soon be taken care of permanently going forward.

dalthviz (approved):
Thanks @luzpaz !

PR #366, opened 2022-09-02
PR: Aliased `QUndoCommand`, provided `QLineEdit.getTextMargins()` and `QFontMetricsF.width()`

### LINKED ISSUE #365: Additions needed for PySide6 support (`QFontMetricsF.width`, `QLineEdit.getTextMargins` and `QtWidgets.QUndoCommand`)
I recently ported pycode to support PySide6, including removing the pyqode.qt compatibility module and using qtpy instead. To complete that, I needed to add some additional support:

```
import os
from qtpy import PYSIDE6_API
qt_api = os.environ['QT_API'].lower()
if qt_api in PYSIDE6_API:
    from PySide6.QtGui import *
    from PySide6.QtWidgets import *
    QFontMetrics.width = lambda self, *args, **kwargs: self.horizontalAdvance(*args, **kwargs)
    QFontMetricsF.width = lambda self, *args, **kwargs: self.horizontalAdvance(*args, **kwargs)
    QLineEdit.getTextMargins = lambda self: (self.textMargins().left(), self.textMargins().top(), self.textMargins().right(), self.textMargins().bottom())
```

Additionally I needed to add in a single source file (I didn't modify my qtpy files at all):

```
try:
    from qtpy import QtGui
    QtWidgets.QUndoCommand = QtGui.QUndoCommand
except:
    pass
```

random-developer:
Fixes #365

### DIFF
--- a/qtpy/tests/test_qtgui.py
+++ b/qtpy/tests/test_qtgui.py
@@ -12,12 +12,16 @@ from qtpy.tests.utils import not_using_conda
     sys.platform.startswith('linux') and not_using_conda(),
     reason="Fatal Python error: Aborted on Linux CI when not using conda")
 def test_qfontmetrics_width(qtbot):
-    """Test QFontMetrics width"""
+    """Test QFontMetrics and QFontMetricsF width"""
     assert QtGui.QFontMetrics.width is not None
+    assert QtGui.QFontMetricsF.width is not None
     font = QtGui.QFont("times", 24)
     font_metrics = QtGui.QFontMetrics(font)
+    font_metricsF = QtGui.QFontMetricsF(font)
     width = font_metrics.width("Test")
+    widthF = font_metricsF.width("Test")
     assert width in range(40, 62)
+    assert widthF == width
 
 
 @pytest.mark.skipif(
--- a/qtpy/tests/test_qtgui.py
+++ b/qtpy/tests/test_qtgui.py
@@ -21,7 +21,7 @@ def test_qfontmetrics_width(qtbot):
     width = font_metrics.width("Test")
     widthF = font_metricsF.width("Test")
     assert width in range(40, 62)
-    assert widthF == width
+    assert 39 <= widthF <= 63
 
 
 @pytest.mark.skipif(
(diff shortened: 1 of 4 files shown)


dalthviz (commented):
Thanks @random-developer for your help with this! I think the only missing thing here is to add some testing:

* For `QtWidgets.QUndoCommand` and `QtWidgets.QLineEdit.getTextMargins` at: https://github.com/spyder-ide/qtpy/blob/master/qtpy/tests/test_qtwidgets.py
* For `QtGui.QFontMetricsF.width` at: https://github.com/spyder-ide/qtpy/blob/master/qtpy/tests/test_qtgui.py similar to https://github.com/spyder-ide/qtpy/blob/fdd6818d58d916e5b144dcf99e3a9e22d23d42d5/qtpy/tests/test_qtgui.py#L14-L20

Other than that this looks good to me :+1:

random-developer:
The QUndoCommand test is failing. I'm not sure why; will investigate later; too pressed for time right now.

qtpy/tests/test_qtgui.py
```diff
@@ -12,12 +12,16 @@
...
+    """Test QFontMetrics and QFontMetricsF width"""
     assert QtGui.QFontMetrics.width is not None
+    assert QtGui.QFontMetricsF.width is not None
     font = QtGui.QFont("times", 24)
     font_metrics = QtGui.QFontMetrics(font)
+    font_metricsF = QtGui.QFontMetricsF(font)
     width = font_metrics.width("Test")
+    widthF = font_metricsF.width("Test")
     assert width in range(40, 62)
+    assert widthF == width
```
dalthviz, line 24: Seems like this is causing the test to file since `widthF` is a floating point number while `width` is an integer so this end up being a validation like `54.484375 == 54` I would suggest to use something like the assert above:

```suggestion
    assert  40 <= widthF <= 60
```

random-developer: Actually that test was passing but your point is solid.

dalthviz (approved):
Thanks @random-developer ! This LGTM :+1:

PR #372, opened 2022-10-09
PR: Add `Qsci` to the imports

### LINKED ISSUE #134: QScintilla Support
I would appear that the Qsci python port in PyQt5 is not supported in QtPy. Is this because it is not supported yet on the other bindings ?

dgoeries:
This adds Qsci to the potential imports for qtpy.

Fixes #134

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,7 +34,7 @@ jobs:
       PYQT6_QT_VERSION: ${{ matrix.pyqt6-qt-version }}
       PYSIDE2_QT_VERSION: ${{ matrix.pyside2-qt-version }}
       PYSIDE6_QT_VERSION: ${{ matrix.pyside6-qt-version }}
-      QSCINTILLA_VERSION: '2.13.*'
+      QSCINTILLA_VERSION: '2.13'
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,6 +34,7 @@ jobs:
       PYQT6_QT_VERSION: ${{ matrix.pyqt6-qt-version }}
       PYSIDE2_QT_VERSION: ${{ matrix.pyside2-qt-version }}
       PYSIDE6_QT_VERSION: ${{ matrix.pyside6-qt-version }}
+      QSCINTILLA_VERSION: '2.13.*'
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/test.sh
+++ b/.github/workflows/test.sh
@@ -27,9 +27,9 @@ if [ "$USE_CONDA" = "Yes" ]; then
 else
 
     if [ "${1}" = "pyqt5" ]; then
-        pip install pyqt5==${PYQT5_VERSION:-"5.15"}.* PyQtWebEngine==${PYQT5_VERSION:-"5.15"}.* QScintilla==$QSCINTILLA_VERSION
+        pip install pyqt5==${PYQT5_VERSION:-"5.15"}.* PyQtWebEngine==${PYQT5_VERSION:-"5.15"}.* QScintilla==${QSCINTILLA_VERSION:-"2.13"}.*
     elif [ "${1}" = "pyqt6" ]; then
-        pip install pyqt6==${PYQT6_VERSION:-"6.3"}.* PyQt6-WebEngine==${PYQT6_VERSION:-"6.3"}.* PyQt6-Qt6==${PYQT6_QT_VERSION:-"6.3"}.* PyQt6-QScintilla==$QSCINTILLA_VERSION
+        pip install pyqt6==${PYQT6_VERSION:-"6.3"}.* PyQt6-WebEngine==${PYQT6_VERSION:-"6.3"}.* PyQt6-Qt6==${PYQT6_QT_VERSION:-"6.3"}.* PyQt6-QScintilla==${QSCINTILLA_VERSION:-"2.13"}.*
     elif [ "${1}" = "pyside2" ]; then
         pip install pyside2==${PYSIDE2_VERSION:-"5.12"}.*
     elif [ "${1}" = "pyside6" ]; then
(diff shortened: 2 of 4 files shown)


dgoeries:
Have to include that in the CI. Some maintainer wants to take this quickly over for the workflow or wants to provide guidance?

qtpy/Qsci.py
```diff
@@ -0,0 +1,28 @@
...
+    PYSIDE2,
+    PYSIDE6,
+    QtBindingsNotFoundError,
+    QtBindingMissingModuleError,
+)
+
+if PYQT5:
+    from PyQt5.Qsci import *
+elif PYQT6:
+    from PyQt6.Qsci import *
```
dalthviz, line 22: I think we should add here some logic to handle the extra package that provides the QSintilla bindings not being available/not installed. So something like:
```suggestion
from . import (
    PYQT5,
    PYQT6,
    PYSIDE2,
    PYSIDE6,
    QtBindingsNotFoundError,
    QtBindingMissingModuleError,
    QtModuleNotInstalledError
)

if PYQT5:
    try:
        from PyQt5.Qsci import *
    except ModuleNotFoundError as error:
        raise QtModuleNotInstalledError(
            name='Qsci', missing_package='QScintilla'
        ) from error
elif PYQT6:
    try:
        from PyQt6.Qsci import *
    except ModuleNotFoundError as error:
        raise QtModuleNotInstalledError(
            name='Qsci', missing_package='PyQt6-QScintilla'
        ) from error
```

dalthviz (commented):
Thank you @dgoeries for the help with this! I left some suggestion to better handle the case when the extra packages required to import `QScintilla` related things are not available, and also some suggestions for the test `skipif` condition.

As a side note, checking the changes done here to add QScintilla to the test suit, maybe we should make the pip jobs install all the extra packages like `QScintilla/PyQt6-QScintilla` (packages besides `PyQtWebEngine` like `PyQtCharts`, `PyQt3D` and others). What do you think @CAM-Gerlach ?

dgoeries:
Thanks a lot @dalthviz !

.github/workflows/test.sh
```diff
@@ -27,9 +27,9 @@ if [ "$USE_CONDA" = "Yes" ]; then
 else
 
     if [ "${1}" = "pyqt5" ]; then
-        pip install pyqt5==${PYQT5_VERSION:-"5.15"}.* PyQtWebEngine==${PYQT5_VERSION:-"5.15"}.*
+        pip install pyqt5==${PYQT5_VERSION:-"5.15"}.* PyQtWebEngine==${PYQT5_VERSION:-"5.15"}.* QScintilla==$QSCINTILLA_VERSION
```
dalthviz, line 30: Lets add a default value for the new `QSCINTILLA_VERSION` value here, so something like:

```suggestion
        pip install pyqt5==${PYQT5_VERSION:-"5.15"}.* PyQtWebEngine==${PYQT5_VERSION:-"5.15"}.* QScintilla==${QSCINTILLA_VERSION:-"2.13"}.*
```

.github/workflows/test.sh
```diff
@@ -27,9 +27,9 @@ if [ "$USE_CONDA" = "Yes" ]; then
...
     elif [ "${1}" = "pyqt6" ]; then
-        pip install pyqt6==${PYQT6_VERSION:-"6.3"}.* PyQt6-WebEngine==${PYQT6_VERSION:-"6.3"}.* PyQt6-Qt6==${PYQT6_QT_VERSION:-"6.3"}.*
+        pip install pyqt6==${PYQT6_VERSION:-"6.3"}.* PyQt6-WebEngine==${PYQT6_VERSION:-"6.3"}.* PyQt6-Qt6==${PYQT6_QT_VERSION:-"6.3"}.* PyQt6-QScintilla==$QSCINTILLA_VERSION
```
dalthviz, line 32: In a similar way as above, lets add a default value here:
```suggestion
        pip install pyqt6==${PYQT6_VERSION:-"6.3"}.* PyQt6-WebEngine==${PYQT6_VERSION:-"6.3"}.* PyQt6-Qt6==${PYQT6_QT_VERSION:-"6.3"}.* PyQt6-QScintilla==${QSCINTILLA_VERSION:-"2.13"}.*
```

.github/workflows/ci.yml
```diff
@@ -34,6 +34,7 @@ jobs:
       PYQT6_QT_VERSION: ${{ matrix.pyqt6-qt-version }}
       PYSIDE2_QT_VERSION: ${{ matrix.pyside2-qt-version }}
       PYSIDE6_QT_VERSION: ${{ matrix.pyside6-qt-version }}
+      QSCINTILLA_VERSION: '2.13.*'
```
dalthviz, line 37: Following the changes suggested below, this will change to be something like:
```suggestion
      QSCINTILLA_VERSION: '2.13'
```

dalthviz (commented):
Thanks for addressing the initial review @dgoeries ! I left a couple more suggestions regarding the CI setup. Other than that this LGTM :+1:

dalthviz (approved):
Thanks @dgoeries for all the work here!

CAM-Gerlach (commented):
LGTM overall, thanks @dgoeries and @dalthviz .

It is a little odd to me to install and test Qscintillla, of all add-ons (being newly added and only on one binding), but not the others. Ideally we'd have a matrix with and without optional addons, but realistically it probably makes the most sense to just test with the various add-on packages, rather than without (especially if we're just skipping the tests instead of checking the correct exception is raised, or at least strict-xfailing them), considering there are more valuable things we would want to spend our matrix on, e.g. more max and min versions.

It might also be a good idea to have an opt-in or opt-out warning of modules that are not available on all supported Qt or binding versions, for users who want to ensure full cross-version/binding compatibility (ala the purpose of QtPy). However, I don't think that's in scope here.

dalthviz:
Thanks for the feedback @CAM-Gerlach ! I will proceed merging this one and in a new PR adding to the test suite setup the installation of other packages :+1:

dgoeries:
Thanks a lot!

Issue #373, opened 2022-10-13
PySide6 6.40 support - `AttributeError: Cannot reassign members` when aliasing enum values


jiayouzl:
Traceback (most recent call last):
  File "/Users/zhanglei/Python代码/PyQT6例程/一键生成pyinstaller编译参数/main.py", line 12, in <module>
    import qtawesome as qta
  File "/opt/homebrew/lib/python3.9/site-packages/qtawesome/__init__.py", line 20, in <module>
    from qtpy import QtCore, QtWidgets, QtGui
  File "/opt/homebrew/lib/python3.9/site-packages/qtpy/QtCore.py", line 116, in <module>
    Qt.MidButton = Qt.MouseButton.MiddleButton = Qt.MiddleButton
  File "shibokensupport/enum_310.py", line 529, in __setattr__
AttributeError: Cannot reassign members.

astrofrog:
I'm having this issue too with PySide6 on different platforms:

```
In [1]: from qtpy.QtCore import Qt

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Input In [1], in <cell line: 1>()
----> 1 from qtpy.QtCore import Qt

File ~/python/dev/lib/python3.10/site-packages/qtpy/QtCore.py:112, in <module>
    110 Qt.BackgroundColorRole = Qt.ItemDataRole.BackgroundColorRole = Qt.BackgroundRole
    111 Qt.TextColorRole = Qt.ItemDataRole.TextColorRole = Qt.ForegroundRole
--> 112 Qt.MidButton = Qt.MouseButton.MiddleButton = Qt.MiddleButton
    114 # Map DeprecationWarning methods
    115 QCoreApplication.exec_ = QCoreApplication.exec

File /Library/Frameworks/Python.framework/Versions/3.10/lib/python3.10/enum.py:480, in EnumMeta.__setattr__(cls, name, value)
    478 member_map = cls.__dict__.get('_member_map_', {})
    479 if name in member_map:
--> 480     raise AttributeError('Cannot reassign members.')
    481 super().__setattr__(name, value)

AttributeError: Cannot reassign members.
```

PR #374, opened 2022-10-16
Don't re-assign Qt.MouseButton.MiddleButton on PySide6

astrofrog:
I think this isn't needed, at least ``Qt.MouseButton.MiddleButton`` seems to exist with PySide6 6.2, 6.3, and 6.4. In 6.4 re-assigning this leads to an actual error.

Fixes https://github.com/spyder-ide/qtpy/issues/373


dalthviz (approved):
Thanks @astrofrog for the help with this! LGTM :+1:

PR #379, opened 2022-10-20
PR: Raise error when no bindings are found at `__init__`

### LINKED ISSUE #367: Import behaviour if no Qt bindings are installed
Hi, firstly thanks very much for QtPy - the flexibility it brings has been very useful for a recent Qt gui I've been working on. While testing that project, I noticed something which I think is unintended behaviour with qtpy.

This concerns only the case where *no Qt bindings are installed*. In a clean venv, with only qtpy, pip, wheel and setuptools installed, with the current version of qtpy, 

```python
>>> import qtpy
>>> qtpy.API_NAME
'PyQt5'
```

importing the package succeeds and the API_NAME is set to "PyQt5", even though there are no Qt bindings installed. Also e.g.:

```python
>>> import qtpy.QtCore
  File "E:\test_qtpy\qtpy\qtpy\QtCore.py", line 15, in <module>
    from PyQt5.QtCore import *
ModuleNotFoundError: No module named 'PyQt5'
```

results in a ModuleNotFoundError because the program is looking for PyQt5, rather than raising a QtBindingsNotFoundError.

- I think that this can be fixed by replacing line 259 in \_\_init\_\_.py with "raise QtBindingsNotFoundError"

- There also appears to be a bug in the definition of QtBindingsNotFoundError: in line 84 in \_\_init\_\_.py, a "self." needs to be inserted before _msg.

With these two changes:

```python
>>> import qtpy
  File "E:\test_qtpy\lib\site-packages\qtpy\__init__.py", line 252, in <module>
ModuleNotFoundError: No module named 'PySide6'

During handling of the above exception, another exception occurred:

  File "E:\test_qtpy\lib\site-packages\qtpy\__init__.py", line 259, in <module>
    raise QtBindingsNotFoundError
qtpy.QtBindingsNotFoundError: No Qt bindings could be found
```

- As \_\_init\_\_.py is executed before importing any other modules, an attempted import of any other module will result in the QtBindingsNotFoundError
- I think this means it may be unnecessary for other modules to import the QtBindingsNotFoundError, as this case would be already handled by \_\_init\_\_.py.

Hope this might be useful. Thanks again.

dalthviz:
Following suggestion at https://github.com/spyder-ide/qtpy/issues/367#issue-1377016517 by @tgwoodcock 

Fixes #367

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -256,7 +256,7 @@ if API in PYSIDE6_API:
         QT6 = PYSIDE6 = True
 
     except ImportError:
-        API = 'pyqt5'
+        raise QtBindingsNotFoundError()
     else:
         os.environ[QT_API] = API
 
--- a/qtpy/Qsci.py
+++ b/qtpy/Qsci.py
@@ -12,7 +12,6 @@ from . import (
     PYQT6,
     PYSIDE2,
     PYSIDE6,
-    QtBindingsNotFoundError,
     QtBindingMissingModuleError,
     QtModuleNotInstalledError
 )
@@ -35,5 +34,3 @@ elif PYSIDE2:
     raise QtBindingMissingModuleError(name='Qsci')
 elif PYSIDE6:
     raise QtBindingMissingModuleError(name='Qsci')
-else:
-    raise QtBindingsNotFoundError()
--- a/qtpy/Qt3DAnimation.py
+++ b/qtpy/Qt3DAnimation.py
@@ -12,7 +12,6 @@ from . import (
     PYQT6,
     PYSIDE2,
     PYSIDE6,
-    QtBindingsNotFoundError,
     QtModuleNotInstalledError,
 )
 
@@ -44,5 +43,3 @@ elif PYSIDE6:
 
     for __name in inspect.getmembers(__temp.Qt3DAnimation):
         globals()[__name[0]] = __name[1]
-else:
-    raise QtBindingsNotFoundError()
--- a/qtpy/Qt3DCore.py
+++ b/qtpy/Qt3DCore.py
@@ -12,7 +12,6 @@ from . import (
     PYQT6,
     PYSIDE2,
     PYSIDE6,
-    QtBindingsNotFoundError,
     QtModuleNotInstalledError,
 )
 
@@ -44,5 +43,3 @@ elif PYSIDE6:
 
     for __name in inspect.getmembers(__temp.Qt3DCore):
         globals()[__name[0]] = __name[1]
-else:
-    raise QtBindingsNotFoundError()
--- a/qtpy/Qt3DExtras.py
+++ b/qtpy/Qt3DExtras.py
@@ -12,7 +12,6 @@ from . import (
     PYQT6,
     PYSIDE2,
     PYSIDE6,
-    QtBindingsNotFoundError,
     QtModuleNotInstalledError,
 )
 
@@ -44,5 +43,3 @@ elif PYSIDE6:
 
     for __name in inspect.getmembers(__temp.Qt3DExtras):
         globals()[__name[0]] = __name[1]
-else:
-    raise QtBindingsNotFoundError()
--- a/qtpy/Qt3DInput.py
+++ b/qtpy/Qt3DInput.py
@@ -12,7 +12,6 @@ from . import (
     PYQT6,
     PYSIDE2,
     PYSIDE6,
-    QtBindingsNotFoundError,
     QtModuleNotInstalledError,
 )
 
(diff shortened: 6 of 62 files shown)


qtpy/__init__.py
```diff
@@ -256,7 +256,7 @@ def __init__(self, *, missing_package=None, **superclass_kwargs):
         QT6 = PYSIDE6 = True
 
     except ImportError:
-        API = 'pyqt5'
+        raise QtBindingsNotFoundError()
```
ccordoba12, line 259: It's not clear to me if `QT_API` is set at this point or not. If it is, then we should clear it here too.

ccordoba12 (commented):
I have a small doubt about this (mentioned below), otherwise looks good to me.

qtpy/__init__.py
dalthviz: At this point `QT_API` is not set as an env variable yet (the `os.environ[QT_API]` call is only done if a successful binding import is done). However, `API` has as value `pyside6` at this point. Should we change the variable `API` value before raising the `QtBindingsNotFoundError` to something like `''` or `None` @ccordoba12 ?

ccordoba12: > At this point QT_API is not set as an env variable yet

Ok, that good to know.

>  Should we change the variable API value before raising the QtBindingsNotFoundError to something like '' or None

I think you could simply remove it from the environment by doing `os.environ.pop(`API`)`.

dalthviz: `API` is just a variable which is initialized with `os.environ.get(QT_API, 'pyqt5')` and that is changed while the different bindings imports are being tried. If a binding import is successful, then the value of `API` is used to set the `QT_API` env var with something like `os.environ[QT_API] = API`. So asically there is no env var `API` set at any point

ccordoba12: Sorry, I thought `API` was also an environment variable. Then I think there's no need to change its value to something else because at this point the package will raise `QtBindingsNotFoundError` and `API` nor `QT_API` will be leaked to the environment (which was my concern).

ccordoba12 (approved):
Great work, thanks @dalthviz!

dalthviz:
Merging this to proceed with the release of v2.3.0

dalthviz:
Note: This is missing and update to take into account the modules added at PR #382

dalthviz:
Merging this to proceed with the release

CAM-Gerlach:
Sorry I didn't get the chance to review this at the time. However, I don't understand why a `QtBindingsNotFoundError` is raised in every submodule, instead of just once in `__init__,py`? Not only is a big code smell and a maintenance burden to add and update this in _every_ module, but the former code should not ever be able to even execute (and at least per my basic testing, doesn't), because the package `__init__.py` executes before any submodule, and even if it didn't, is explicitly imported in every submodule (as the main constants and even `QtBindingsNotFoundError` itself is defined there), so the error is raised in `__init__.py` before that line can ever possibly execute. And not only that, even if the line was somehow reached, it _still_ wouldn't fire because the one of the `PYQT5`, `PYSIDE2`, etc. constants (usually `PYQT5`) would still be true even if importing `__init__.py` somehow succeeded without an error, so that block still could not be reached. Is there something big I'm missing here?

Also, the correct fix (using `raise QtBindingsNotFoundError from None`) we discussed on the issue that solves the two problems I pointed out with the initial naive implementation were not implemented either, making the error/traceback message and behavior less appropriate and more confusing, as it is treated and printed as an unexpected error during error handling rather than a intentional wrapper error (with `raise from`), and furthermore prints the error message and traceback from importing `pyside6`, making users think it was incorrectly trying to use that binding instead of the one they selected.

Here's how the error output looks now (note that even though `QtCore` is imported, the `__init__.py` error is still raised anyway and the core in the module is never touched):

```python
>>> import qtpy.QtCore
  File "C:\Users\C. A. M. Gerlach\Documents\dev\SpyderDev\qtpy\qtpy\__init__.py", line 252, in <module>
ModuleNotFoundError: No module named 'PySide6'

During handling of the above exception, another exception occurred:

  File "C:\Users\C. A. M. Gerlach\Documents\dev\SpyderDev\qtpy\qtpy\__init__.py", line 259, in <module>
    raise QtBindingsNotFoundError()
qtpy.QtBindingsNotFoundError: No Qt bindings could be found
```

and here's how it should look with the correct `raise QtBindingsNotFoundError from None`:

```python
>>> import qtpy.QtCore
  File "C:\Users\C. A. M. Gerlach\Documents\dev\SpyderDev\qtpy\qtpy\__init__.py", line 259, in <module>
    raise QtBindingsNotFoundError from None
qtpy.QtBindingsNotFoundError: No Qt bindings could be found
```

CAM-Gerlach:
I've opened issue #390 to document this, and PR #391 to fix this.

PR #380, opened 2022-10-21
PR: Enable more qt6 tests

jschueller:
the disabled classes QCameraViewfinder, QSGAbstractRenderer& QSGEngine are qt5-only

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,7 +79,7 @@ jobs:
         shell: bash
         run: |
           sudo apt update
-          sudo apt install libpulse-dev libegl1-mesa libopengl0
+          sudo apt install libpulse-dev libegl1-mesa libopengl0 gstreamer1.0-gl
       - name: Install Conda
         uses: conda-incubator/setup-miniconda@v2
         with:
--- a/qtpy/tests/test_qtmultimediawidgets.py
+++ b/qtpy/tests/test_qtmultimediawidgets.py
@@ -6,15 +6,12 @@ from qtpy import PYQT5, PYSIDE2
 from qtpy.tests.utils import using_conda
 
 
-@pytest.mark.skipif(not (PYQT5 or PYSIDE2), reason="Only available in Qt5 bindings")
-@pytest.mark.skipif(
-    using_conda(), reason="Conda packages don't seem to include QtMultimedia"
-)
 def test_qtmultimediawidgets():
     """Test the qtpy.QtMultimediaWidgets namespace"""
     from qtpy import QtMultimediaWidgets
 
-    assert QtMultimediaWidgets.QCameraViewfinder is not None
+    if PYQT5 or PYSIDE2:
+        assert QtMultimediaWidgets.QCameraViewfinder is not None
+        # assert QtMultimediaWidgets.QVideoWidgetControl is not None
     assert QtMultimediaWidgets.QGraphicsVideoItem is not None
     assert QtMultimediaWidgets.QVideoWidget is not None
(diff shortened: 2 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -79,7 +79,7 @@ jobs:
         shell: bash
         run: |
           sudo apt update
-          sudo apt install libpulse-dev libegl1-mesa libopengl0
+          sudo apt install libpulse-dev libegl1-mesa libopengl0 gstreamer1.0-plugins-base gstreamer1.0-gl
```
dalthviz, line 82: I think this is needed for `test_multimedia.py` but I think no changes have been done to that test file. Maybe this is needed for some other test? If not then I think we should revert this change

qtpy/tests/test_qtmultimediawidgets.py
```diff
@@ -6,15 +6,11 @@
...
-@pytest.mark.skipif(not (PYQT5 or PYSIDE2), reason="Only available in Qt5 bindings")
-@pytest.mark.skipif(
-    using_conda(), reason="Conda packages don't seem to include QtMultimedia"
-)
 def test_qtmultimediawidgets():
     """Test the qtpy.QtMultimediaWidgets namespace"""
     from qtpy import QtMultimediaWidgets
 
-    assert QtMultimediaWidgets.QCameraViewfinder is not None
+    # assert QtMultimediaWidgets.QCameraViewfinder is not None
```
dalthviz, line 13: I think it would be better to add a validation to test this instead of just commenting the assert, something like:

```python
if QT5:    # Only available for Qt5 bindings
    assert QtMultimediaWidgets.QCameraViewfinder is not None
```
Probably the `QtMultimediaWidgets.QVideoWidgetControl` was commented for that reason and we could move it to the Qt5 validation?

qtpy/tests/test_qtquick.py
```diff
@@ -32,11 +31,11 @@ def test_qtquick():
     assert QtQuick.QQuickTextureFactory is not None
     assert QtQuick.QQuickView is not None
     assert QtQuick.QQuickWindow is not None
-    assert QtQuick.QSGAbstractRenderer is not None
+    # assert QtQuick.QSGAbstractRenderer is not None
     assert QtQuick.QSGBasicGeometryNode is not None
     assert QtQuick.QSGClipNode is not None
     assert QtQuick.QSGDynamicTexture is not None
-    assert QtQuick.QSGEngine is not None
+    # assert QtQuick.QSGEngine is not None
```
dalthviz, line 38: I think it would be better to add here too a validation to test these modules instead of just commenting the asserts, something like:

```python
if QT5:    # Only available for Qt5 bindings
    assert QtQuick.QSGAbstractRenderer is not None
    assert QtQuick.QSGEngine is not None
```

dalthviz (commented):
Thanks @jschueller for working on this! I left some comments/suggestions regarding the commented asserts for modules only available on Qt5 bindings. Also, left a question regarding the installation of GStreamer for Linux.

.github/workflows/ci.yml
jschueller: this is needed for the pyqt6 wheels

qtpy/tests/test_qtquick.py
jschueller: ok

qtpy/tests/test_qtmultimediawidgets.py
jschueller: actually QVideoWidgetControl seems not available with pyqt5

dalthviz (approved):
Thanks @jschueller !

PR #382, opened 2022-10-27
PR: Add QtPdf

### LINKED ISSUE #381: Add QtPdf, QtPdfWidgets
these are new in 6.x

jschueller:
Closes #381

### DIFF
--- a/qtpy/QtPdf.py
+++ b/qtpy/QtPdf.py
@@ -1,0 +1,28 @@
+# -----------------------------------------------------------------------------
+# Copyright © 2009- The Spyder Development Team
+#
+# Licensed under the terms of the MIT License
+# (see LICENSE.txt for details)
+# -----------------------------------------------------------------------------
+
+"""Provides QtPdf classes and functions."""
+
+from . import (
+    PYQT5,
+    PYQT6,
+    PYSIDE2,
+    PYSIDE6,
+    QtBindingsNotFoundError,
+    QtBindingMissingModuleError,
+)
+
+if PYQT5:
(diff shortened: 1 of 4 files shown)


dalthviz (commented):
Thanks @jschueller for helping with this! Checking, I think that the `QtPdf` and `QtPdfWidgets` modules are only available in PyQt6 and PySide6 >=6.4.0. Taking that into account, maybe we should put in place a try-catch for the PyQt6 and PySide6 import calls and raise a `QtBindingMissingModuleError` in case of an `ImportError`? What do you think?

Beside that, could you rebase on top of the latest master, please? That should fix the failing test.

Other than the above this LGTM :+1:

jschueller:
ok, I rebased
I think ImportError is fine, what does QtBindingMissingModuleError bring to the table ?

dalthviz:
Thanks for the quick response! And regarding raising `QtBindingMissingModuleError` is just to be consistent with other imports that use the specific error classes that QtPy implements and also to show that is expected for the module import to fail in some circumstances (in this case if PyQt6 or PySide6 >=6.4)

jschueller:
thtat's not what is done for other modules, let's keep it simple

dalthviz (approved):
Thanks again @jschueller for your help with this! Thinking a little bit more, there is no need to hold this for being merged. Indeed, keeping things simple is better, and actually there is no qtpy error implemented that encapsulates the case of a module not being available in a specific binding version (maybe in the future we could add it but for the moment the `ModuleNotFoundError` should be fine 👍🏼 )

Issue #383, opened 2022-11-01
Add extra packages available for PyQt to CI for testing


dalthviz:
Follow-up issue as discussed at https://github.com/spyder-ide/qtpy/pull/372 to add to the test suite setup extra packages

PR #387, opened 2022-12-10
PR: Improve import modularity between `QtGui`, `QtWidgets` and `QtOpenGL*` related modules

DaelonSuzuka:
For some reason, `QtGui` and `QtWidgets` are importing `QtOpenGL` and `QtOpenGLWidgets`, respectively. Attempting to import these here forces applications to depend on OpenGL even if they don't need it.

My use case for this change is that I distribute Qt/Python applications by creating a single-folder bundle with PyInstaller, and then zipping that bundle to create a portable distribution, or compiling it into an installer with a tool like InnoSetup. I was looking to reduce the size of my distributions, and Qt is supposed to be able to run with only a few of the dlls present, so I started deleting the biggest ones from my bundle. `opengl32sw.dll` is 20MB and `QtOpenGL.pyd` is 10MB, so omitting them is a non-trivial reduction in final size.

Am I missing a reason for these imports?

If removing these is okay, I can also audit the rest of the package for similar cross-contaminations (and I promise it won't spiral into a multi-month project like last time >.>).


ccordoba12:
> However, maybe we could somehow condition making these imports with a flag (something like QTOPENGL_QT5_LAYOUT)? What do you think @CAM-Gerlach @ccordoba12 ?

Good idea, I was thinking along those lines too.

CAM-Gerlach:
I guess that seems reasonable to me, a bit ugly but I'm not sure there's really another alternative, other than making it a lazy import or breaking backward compatibility.

If we add a flag, we'd want it to documented in the README like the others. For backward compat, it should presumably by on by default in QtPy 2.x, though we could revisit that in QtPy 3.x.

CAM-Gerlach:
Thinking about it more, ISTM both the simplest and best approach here is try/except the `QtOpenGL` import and if it fails, define a module-level `__getattr__` (supported in Python 3.7+) that raises a `QtModuleNotInstalledError` with the package name (instead of a `AttributeError`) if any of the missing `QtOpenGL` attributes are accessed. i.e. something like:

```python
MISSING_OPENGL_NAMES = {}
# ...
if PYQT6:
    # ...
    try:
        from PyQt6.QtOpenGLWidgets import QOpenGLWidget
    except ModuleNotFoundError:
        MISSING_OPENGL_NAMES.update("QOpenGLWidget")
# ...
def __getattr__(name):
    if name in MISSING_OPENGL_NAMES:
        raise QtModuleNotInstalledError(
            name="PyQt6.QtOpenGLWidgets", missing_package="pyopengl")
    else:
        raise AttributeError(f"module {__name__!r} has no attribute {name!r}")
```

That way, it would work gracefully right now without a hacky manual flag, while immediately raising a clearer and more specific/helpful error if it is not installed. The same approach could be applied to other optional modules as well, where they don't coincide exactly with a top-level QtPy (i.e. PyQt, PySide) submodule. Plus, it's shorter, simpler and can easily be re-used and extended to handle other missing optional modules.

We could also use the same strategy (with a single `__getattr__` taking some data structure of the various things supported on different bindings/versions and raising the appropriate helpful QtPy errors if they are not found, rather than just a generic Attribute Error. That way we can easily handle and maintain a lot of cases without a bunch of bespoke code, and in a declarative and machine-readable way, which could potentially even be supported by static type checkers using our existing mechanism.

CAM-Gerlach:
> Wow, I did not know that was an option.

Yeah; see [PEP 562](https://peps.python.org/pep-0562/) for more details.

> I'll definitely try this out today.

:+1:

qtpy/QtGui.py
```diff
@@ -8,14 +8,27 @@
...
     from PyQt6.QtGui import *
-    from PyQt6.QtOpenGL import *
+
+    # Attempt to import QOpenGLWidget, but if that fails, don't throw an exception until the
+    # symbol is explicitly asked for
+    # this allows removing the sizeable OpenGL binaries from pyinstaller bundles, but
+    # maintains compatibility with apps that expect this symbol to be automatically imported
+    try:
+        from PyQt6.QtOpenGLWidgets import QOpenGLWidget
+    except (ModuleNotFoundError, ImportError):
```
DaelonSuzuka, line 29: A missing DLL throws an `ImportError`, so I'm catching both error types here.

qtpy/QtGui.py
```diff
@@ -35,7 +48,16 @@
...
-    from PySide6.QtOpenGL import *
+
+    # Attempt to import QOpenGLWidget, but if that fails, don't throw an exception until the
+    # symbol is explicitly asked for
+    # this allows removing the sizeable OpenGL binaries from pyinstaller bundles, but
+    # maintains compatibility with apps that expect this symbol to be automatically imported
+    try:
+        from PySide6.QtOpenGLWidgets import QOpenGLWidget
+    except (ModuleNotFoundError, ImportError):
+        MISSING_OPENGL_NAMES['QOpenGLWidget'] = ('PySide6.QtOpenGLWidgets', 'pyopengl')
```
DaelonSuzuka, line 59: The definition of the error message contents belongs closer to where the error happens.

qtpy/QtGui.py
```diff
@@ -72,3 +94,11 @@ def movePositionPatched(
     ) -> bool:
         return movePosition(self, operation, mode, n)
     QTextCursor.movePosition = movePositionPatched
+
+
+def __getattr__(name):
+    if name in MISSING_OPENGL_NAMES:
+        module, package = MISSING_OPENGL_NAMES[name]
+        raise QtModuleNotInstalledError(name=module, missing_package=package)
```
DaelonSuzuka, line 102: This is now totally generic, since the message is defined above.

qtpy/QtGui.py
```diff
@@ -8,14 +8,27 @@
 
 """Provides QtGui classes and functions."""
 
-from . import PYQT6, PYQT5, PYSIDE2, PYSIDE6
+from . import PYQT6, PYQT5, PYSIDE2, PYSIDE6, QtModuleNotInstalledError
+
+
+MISSING_OPENGL_NAMES = {}
```
DaelonSuzuka, line 14: Is `MISSING_OPENGL_NAMES` still the right name for this variable? This mechanism should be able to handle deferring the error from any imports, not just OpenGL stuff.

I did briefly look for other modules that I might want to strip for size and didn't see any. I think this is only an issue if `qtpy` crosses Qt module boundaries, and everything else I saw was already properly contained.

qtpy/QtWidgets.py
```diff
@@ -8,15 +8,27 @@
...
-    from PyQt6.QtOpenGLWidgets import QOpenGLWidget
+
+    # Attempt to import QOpenGLWidget, but if that fails, don't throw an exception until the
+    # symbol is explicitly asked for
+    # this allows removing the sizeable OpenGL binaries from pyinstaller bundles, but
+    # maintains compatibility with apps that expect this symbol to be automatically imported
+    try:
+        from PyQt6.QtOpenGLWidgets import QOpenGLWidget
+    except (ModuleNotFoundError, ImportError):
+        MISSING_SYMBOL_ERRORS['QOpenGLWidget'] = QtModuleNotInstalledError(name='PyQt6.QtOpenGLWidgets', missing_package='pyopengl')
```
DaelonSuzuka, line 31: This alternate style puts the error site in _complete_ control of what gets raised, instead of just the message contents.

I think I prefer this because all the 'thinking' is up here, and the `__getattr__` is now purely mechanical.

It's possible that the (potential) reuse of an exception instance is dangerous somehow; I couldn't find an answer to this online. Making this a lambda should be a sufficient workaround, if necessary.

DaelonSuzuka:
Sorry for the conversation spam, I had to made some modifications to the mechanism and I wanted to explain the rationale.

`QtGui.py` and `QtWidgets.py` have slightly different implementations because I thought of a possible refactor while I was writing the first set of comments. I changed only `QtWidgets.py` to get feedback on both versions. The winning style will replace the losing one.

# The Real Problem

How on earth do I write tests for this? I manually tested it in a project of mine by:
- uninstall qtpy
- install qtpy from local repo
- delete `build/` and `dist/`
- run pyinstaller to make a single-folder bundle
- run bundled exe to make sure it works
- open `project/dist/Application/PySide6`
- delete the OpenGL files
- run bundled exe **again** to make sure it _still_ works

Something tells me there isn't a pytest helper for this.

CAM-Gerlach (commented):
Thanks for implementing this, @DaelonSuzuka !

I didn't think about it before, but we need to set `__cause__` on the error, so we can:

* Actually indicate that that it was an attribute access that was the proximate cause of the error, and inform the user which attribute required PyOpenGL
* Give the details of the ImportError/ModuleNotFound error that happened on import, giving them useful information about what actually happened rather than leaving them (and us or whomever they report it to) guessing whether the module is missing, there's a DLL issue, it was installed/uninstalled incorrectly, it was pip/conda mixing, etc.

Therefore, after a bunch of experimentation, since this added a lot of logic to every call site, I ended up putting this in a helper utility function and and calling that in `__getattr__`.

I was originally going to make this as suggestions, but it turned out to be rather complicated, and prevent putting the common logic into a utility function somewhere else. Thus, I ended up just pushing it to [a branch on my fork](https://github.com/CAM-Gerlach/qtpy/tree/improve-import-hygiene) you can pull with

```bash
git remote add cam-gerlach https://github.com/CAM-Gerlach/qtpy.git
git fetch cam-gerlach
git pull cam-gerlach improve-import-hygiene
```

Or, I can push it to your branch directly, if you prefer. Thanks!

CAM-Gerlach:
Seems like there's a merge conflict to solve, but we can deal with it after the current batch of changes (to avoid invalidating them).

> Sorry for the conversation spam, I had to made some modifications to the mechanism and I wanted to explain the rationale.

NB, if you want to avoid others getting with with notifications for every comment, you can just do the comments as part of a review of your own PR, which will mark the as pending and then only send one notification once you submit it.

> `QtGui.py` and `QtWidgets.py` have slightly different implementations because I thought of a possible refactor while I was writing the first set of comments. I changed only `QtWidgets.py` to get feedback on both versions. The winning style will replace the losing one.

Ah, I see. I preferred your second, but I ended up refactoring it further and posted a patch/branch with it.

> How on earth do I write tests for this? I manually tested it in a project of mine by:

Our eventual goal is to have CI test environments with and without the optional deps, see #383 , which would allow testing this fully integrated, but for now to test this you should be able to patch `PyQt6`/`PySide6`.`QtOpenGLWidgets` with a `Mock` to always raise `ImportError`, to ensure it doesn't do so, and test that the behavior is correct in each case.

qtpy/QtGui.py
```diff
@@ -8,14 +8,27 @@
...
+
 
 if PYQT5:
     from PyQt5.QtGui import *
 elif PYQT6:
     from PyQt6 import QtGui
     from PyQt6.QtGui import *
-    from PyQt6.QtOpenGL import *
```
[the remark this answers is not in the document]
DaelonSuzuka: Sloppy copy and paste error, thanks for catching that!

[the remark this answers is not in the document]
DaelonSuzuka: > you'll need to list the attributes present in PyQt/PySide's QtOpenGL 5.15 explicitly here.

Ah crap, that's much less fun... `qtpy/QtOpenGL.py` has a list of attributes it's importing from `QtGui`, so is it safe to assume this is the list of names we need to support?

This is kinda rough:

```py
try:
    from PySide6.QtOpenGL import *
except ImportError as error:
    names = [
        'QOpenGLBuffer',
        'QOpenGLFramebufferObject',
        'QOpenGLFramebufferObjectFormat',
        'QOpenGLShader',
        'QOpenGLShaderProgram',
        'QOpenGLContext',
        'QOpenGLContextGroup',
        'QOpenGLDebugLogger',
        'QOpenGLDebugMessage',
        'QOpenGLPixelTransferOptions',
        'QOpenGLTexture',
        'QOpenGLTextureBlitter',
        'QOpenGLVersionProfile',
        'QOpenGLVertexArrayObject',
        'QOpenGLWindow',
    ]
    for name in names:
        _missing_optional_names[name] = {
            'name': 'PySide6.QtOpenGL',
            'missing_package': 'pyopengl',
            'import_error': error,
        }
```

DaelonSuzuka:
> you can just do the comments as part of a review of your own PR

Maybe next time. I was a just happy that I managed to get the comments on the right lines...

> Our eventual goal is to have CI test environments with and without the optional deps, see https://github.com/spyder-ide/qtpy/issues/383 , which would allow testing this fully integrated

Not it!

> but for now to test this you should be able to patch PyQt6/PySide6.QtOpenGLWidgets with a Mock to always raise ImportError, to ensure it doesn't do so, and test that the behavior is correct in each case.

I _think_ I know how to do that. Hopefully I can get that done today.

qtpy/QtGui.py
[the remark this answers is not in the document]
CAM-Gerlach: Seems reasonable to me, though not the expert. If those lists are the same, instead of duplicating them maybe just stick them in a private module-level constant named, e.g. `_QTOPENGL_NAMES`, etc. and reference them in both?

qtpy/QtGui.py
```diff
@@ -14,6 +14,24 @@
...
+    'QOpenGLContextGroup',
+    'QOpenGLDebugLogger',
+    'QOpenGLDebugMessage',
+    'QOpenGLPixelTransferOptions',
+    'QOpenGLTexture',
+    'QOpenGLTextureBlitter',
+    'QOpenGLVersionProfile',
+    'QOpenGLVertexArrayObject',
+    'QOpenGLWindow',
+]
```
CAM-Gerlach, line 33:
```suggestion
_QTOPENGL_NAMES = {
    'QOpenGLBuffer',
    'QOpenGLFramebufferObject',
    'QOpenGLFramebufferObjectFormat',
    'QOpenGLShader',
    'QOpenGLShaderProgram',
    'QOpenGLContext',
    'QOpenGLContextGroup',
    'QOpenGLDebugLogger',
    'QOpenGLDebugMessage',
    'QOpenGLPixelTransferOptions',
    'QOpenGLTexture',
    'QOpenGLTextureBlitter',
    'QOpenGLVersionProfile',
    'QOpenGLVertexArrayObject',
    'QOpenGLWindow',
}
```

Also, tiny nit but semantically and efficiently this makes most sense as a set.

qtpy/QtGui.py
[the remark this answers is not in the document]
DaelonSuzuka: `_QTOPENGL_NAMES` is a good idea, consider it done.

I suppose I should audit `PyQt6.QtOpenGL` and `PySide6.QtOpenGL` and compare the available classes to this list.

qtpy/QtGui.py
DaelonSuzuka: Works for me.

DaelonSuzuka:
I don't understand the error on the failed job at all. Is it an environment failure?

ccordoba12:
> I don't understand the error on the failed job at all. Is it an environment failure?

It was just a transient failure, i.e. it was fixed by restarting the CI job.

qtpy/QtGui.py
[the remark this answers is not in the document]
DaelonSuzuka: So I scraped `PyQt6/QtOpenGL.pyi` and `PySide6/QtOpenGL.pyi` and collected all the classes that start with `QOpenGL`. They both had some number of `QOpenGLFunctions_xxxx` classes, whichthat I discarded. 

## Comparison of the QtOpenGL classes to the list in `qtpy/QtOpenGL.py`:

<table>
<tr>
	<th>QtOpenGL.pyi
	<th>qtpy/QtOpenGL.py
<tr>
    <td>QOpenGLBuffer
    <td>QOpenGLBuffer
<tr>
    <td>
    <td>QOpenGLContext
<tr>
    <td>
    <td>QOpenGLContextGroup
<tr>
    <td>QOpenGLDebugLogger
    <td>QOpenGLDebugLogger
<tr>
    <td>QOpenGLDebugMessage
    <td>QOpenGLDebugMessage
<tr>
    <td>QOpenGLFramebufferObject
    <td>QOpenGLFramebufferObject
<tr>
    <td>QOpenGLFramebufferObjectFormat
    <td>QOpenGLFramebufferObjectFormat
<tr>
    <td>QOpenGLPaintDevice
    <td>
<tr>
    <td>QOpenGLPixelTransferOptions
    <td>QOpenGLPixelTransferOptions
<tr>
    <td>QOpenGLShader
    <td>QOpenGLShader
<tr>
    <td>QOpenGLShaderProgram
    <td>QOpenGLShaderProgram
<tr>
    <td>QOpenGLTexture
    <td>QOpenGLTexture
<tr>
    <td>QOpenGLTextureBlitter
    <td>QOpenGLTextureBlitter
<tr>
    <td>QOpenGLTimeMonitor
    <td>
<tr>
    <td>QOpenGLTimerQuery
    <td>
<tr>
    <td>QOpenGLVersionFunctionsFactory
    <td>
<tr>
    <td>QOpenGLVersionProfile
    <td>QOpenGLVersionProfile
<tr>
    <td>QOpenGLVertexArrayObject
    <td>QOpenGLVertexArrayObject
<tr>
    <td>QOpenGLWindow
    <td>QOpenGLWindow
</table>

DaelonSuzuka:
> I think I know how to do that. Hopefully I can get it done today.

Narrator: He did not, in fact, know how to do that.

I'll have to try again after a sleep.

DaelonSuzuka:
Alright, I have to admit defeat: I cannot figure out how to test this. I thought I had it working for PySide6 with monkeypatching, but PyQt6 structures their imports differently and the same technique doesn't do anything there. It also only works if it's the first test that's run, which is obviously not correct.

@CAM-Gerlach pls halp.

CAM-Gerlach:
Okay, sorry about that!

While I previously thought it was possible, at least in some cases, as far as I can tell there's no obvious way to either get the Conda or the PyPI packages without OpenGL, as it is a core part of even the cut-down `PySide6-Essentials` package (aside from, of course, horrible manual hackery as you describe) which, though, does somewhat limit the usefulness of this here for this particular case, though the same code/technique will certainly be useful for others that are more "optional".

For starters, one thing that we should certainly add regardless here is add tests for the OpenGL classes in QtGui and  QtWidgets, to actually verify that we're compatible here, and (that would presumably xfail if OpenGL was not present)—I assuming that's something you've done already locally.

Beyond that, one thing we _could_ do is just add a special matrix job where after installing but before running the test suite, we delete the OpenGL files from the installed package e.g.

```python
for file in Path(PySide6.__file__).parent.glob(*OpenGL*):
    file.unlink()
```

Then run the test suite, or a specific set of tests via a custom marker, checking an environment variable to xfail (with the correct error) or xpass the tests that should fail/pass based on the OpenGL install status.

I'm not sure if that complexity (and an extra matrix job) is worth it for what seems to be a niche case; we have a plan to have separate jobs for scenarios with all the optional deps installed, and only the essentials, to test that things still work as expected, which would would presumably focus on what can be installed/uninstalled normally via pip/Conda, but would at least test the basic logic here.

@dalthviz , what do you think here?

DaelonSuzuka:
> For starters, one thing that we should certainly add regardless here is add tests for the OpenGL classes in QtGui and QtWidgets,

This, I can definitely do.

> Beyond that, one thing we could do is just add a special matrix job where after installing but before running the test suite, we delete the OpenGL files from the installed package e.g.

This was the only idea I had left, but I couldn't tell if it was crazy or not.

dalthviz:
> add tests for the OpenGL classes in QtGui and QtWidgets, to actually verify that we're compatible here, and (that would presumably xfail if OpenGL was not present)

I think that should be enough :+1:

>  Beyond that, one thing we could do is just add a special matrix job where after installing but before running the test suite, we delete the OpenGL files from the installed package e.g.

As you said it seems complex and covers quite a niche case where basically you are removing things manually from the base bindings package, right?

Maybe what could be worthy is to add some tests for the utility functions added and maybe add a test checking that the approach implemented with `QtOpenGL` to handle stuff missing works? So not directly testing `QtOpenGL` but a dummy module only available for the test like `DummyOpenGL` or something like that?

CAM-Gerlach:
> As you said it seems complex and covers quite a niche case where basically you are removing things manually from the base bindings package, right?

That's what it kinda seems like to me, yeah...

> Maybe what could be worthy is to add some tests for the utility functions added and maybe add a test checking that the approach implemented with `QtOpenGL` to handle stuff missing works? So not directly testing `QtOpenGL` but a dummy module only available for the test like `DummyOpenGL` or something like that?

:+1:

DaelonSuzuka:
I _think_ the new tests are working correctly. Thank you @dalthviz for the recommendation.

I tested this new version on one of my work applications and I can strip `opengl32sw.dll`, `QtOpenGL.pyd`, and `Qt6OpenGL.dll` without issue.

qtpy/tests/test_missing_optional_deps.py
```diff
@@ -0,0 +1,20 @@
+"""Test the deferred import error mechanism"""
+
+
+# See https://github.com/spyder-ide/qtpy/pull/387/
+
+
+import pytest
```
CAM-Gerlach, line 7:
```suggestion
import pytest

from qtpy import QtModuleNotInstalledError
```

Needed below

qtpy/tests/test_missing_optional_deps.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def test_missing_optional_deps():
+    """Test importing a module that uses the deferred import error mechanism"""
+    from . import optional_deps
+
+    assert optional_deps.ExampleClass is not None
+
+    with pytest.raises(ImportError) as excinfo:
```
CAM-Gerlach, line 16:
```suggestion
    with pytest.raises(QtModuleNotInstalledError) as excinfo:
```

We should check the actual error class here, no?

qtpy/tests/test_qtwidgets.py
```diff
@@ -4,7 +4,7 @@
 
 import pytest
 
-from qtpy import PYQT5, PYQT_VERSION, QtCore, QtGui, QtWidgets
+from qtpy import PYQT5, PYSIDE2, PYQT_VERSION, QtCore, QtGui, QtWidgets
```
CAM-Gerlach, line 7:
```suggestion
from qtpy import PYQT5, PYQT_VERSION, QtCore, QtGui, QtWidgets
```

I don't see where this is used, sorry—was it needed earlier but later removed?

qtpy/tests/test_qtgui.py
```diff
@@ -123,3 +123,26 @@ def test_qtextcursor_moveposition():
     assert cursor.position() == cursor.anchor()
     assert cursor.movePosition(QtGui.QTextCursor.NextWord, QtGui.QTextCursor.KeepAnchor, 3)
     assert cursor.selectedText() == "foo bar baz"
+
+
+def test_opengl_imports():
+    """Test for presence of QOpenGL* classes
+    These classes were members of QtGui in Qt5, but moved to QtOpenGL in Qt6.
+    QtPy makes them available in QtGui to maintain compatibility
```
CAM-Gerlach, line 131:
```suggestion
    """
    Test for presence of QOpenGL* classes.

    These classes were members of QtGui in Qt5, but moved to QtOpenGL in Qt6.
    QtPy makes them available in QtGui to maintain compatibility.
```

Format the docstring properly (per PEP 257) and consistently

qtpy/tests/test_qtwidgets.py
```diff
@@ -116,3 +116,11 @@ def test_enum_access():
     assert QtWidgets.QStyle.State_None == QtWidgets.QStyle.StateFlag.State_None
     assert QtWidgets.QSlider.TicksLeft == QtWidgets.QSlider.TickPosition.TicksAbove
     assert QtWidgets.QStyle.SC_SliderGroove == QtWidgets.QStyle.SubControl.SC_SliderGroove
+
+
+def test_opengl_imports():
+    """Test for presence of QOpenGLWidget
+    QOpenGLWidget was a member of QtWidgets in Qt5, but moved to QtOpenGLWidgets in Qt6.
+    QtPy makes QOpenGLWidget available in QtWidgets to maintain compatibility
```
CAM-Gerlach, line 124:
```suggestion
    """
    Test for presence of QOpenGLWidget.

    QOpenGLWidget was a member of QtWidgets in Qt5, but moved to QtOpenGLWidgets in Qt6.
    QtPy makes QOpenGLWidget available in QtWidgets to maintain compatibility.
```

Likewise

CAM-Gerlach (commented):
Looking pretty good so far; just some relatively minor suggestions. Thanks!

DaelonSuzuka:
That's the first merge conflict I've resolved using GitHub's editor, I'm crossing my fingers that I didn't mess it up.

qtpy/QtGui.py
```diff
@@ -162,6 +162,19 @@ def movePositionPatched(
     QMouseEvent.globalPos = lambda self: self.globalPosition().toPoint()
     QMouseEvent.globalX = lambda self: self.globalPosition().toPoint().x()
     QMouseEvent.globalY = lambda self: self.globalPosition().toPoint().y()
```
CAM-Gerlach: NB, these and those above (that I can't reach) should be removed in the merge, as they were replaced by the block added below.

CAM-Gerlach (commented):
LGTM except that it appears that the merge conflict was not actually resolved correctly; see comment.

qtpy/QtGui.py
```diff
@@ -108,6 +156,12 @@ def movePositionPatched(
     QMouseEvent.globalPosition = lambda self: __QPointF(
         float(self.globalX()), float(self.globalY()))
 if PYQT6 or PYSIDE6:
+    QMouseEvent.pos = lambda self: self.position().toPoint()
+    QMouseEvent.x = lambda self: self.position().toPoint().x()
+    QMouseEvent.y = lambda self: self.position().toPoint().y()
+    QMouseEvent.globalPos = lambda self: self.globalPosition().toPoint()
+    QMouseEvent.globalX = lambda self: self.globalPosition().toPoint().x()
+    QMouseEvent.globalY = lambda self: self.globalPosition().toPoint().y()
```
CAM-Gerlach, line 164:
```suggestion
```

This block appears to have been accidentally retained in the merge, as it duplicates the added block below.

CAM-Gerlach (commented):
LGTM, except for a problem with the merge resolution (see comment).

CAM-Gerlach (approved):
LGTM from my end, thanks @DaelonSuzuka !

qtpy/QtGui.py
```diff
@@ -8,7 +8,30 @@
...
+    'QOpenGLPixelTransferOptions',
+    'QOpenGLShader',
+    'QOpenGLShaderProgram',
+    'QOpenGLTexture',
+    'QOpenGLTextureBlitter',
+    'QOpenGLVersionProfile',
+    'QOpenGLVertexArrayObject',
+    'QOpenGLWindow',
+}
+
```
ccordoba12, line 34:
```suggestion

def __getattr__(name):
    """Custom getattr to chain and wrap errors due to missing optional deps."""
    raise getattr_missing_optional_dep(
        name, module_name=__name__, optional_names=_missing_optional_names)

```

I moved this function here because I think it'd be easier to spot for future maintenance than if it's placed at the bottom of the file.

qtpy/QtGui.py
```diff
@@ -121,3 +169,8 @@ def movePositionPatched(
     QSinglePointEvent.globalPos = lambda self: self.globalPosition().toPoint()
     QSinglePointEvent.globalX = lambda self: self.globalPosition().toPoint().x()
     QSinglePointEvent.globalY = lambda self: self.globalPosition().toPoint().y()
+
+def __getattr__(name):
+    """Custom getattr to chain and wrap errors due to missing optional deps."""
+    raise _getattr_missing_optional_dep(
+        name, module_name=__name__, optional_names=_missing_optional_names)
```
ccordoba12, line 176:
```suggestion
```

Moved to the top.

qtpy/QtWidgets.py
```diff
@@ -8,15 +8,31 @@
 
 """Provides widget classes and functions."""
 
-from . import PYQT5, PYQT6, PYSIDE2, PYSIDE6
+from . import PYQT5, PYQT6, PYSIDE2, PYSIDE6, QtModuleNotInstalledError
+from .utils import _getattr_missing_optional_dep
+
+
+_missing_optional_names = {}
+
```
ccordoba12, line 16:
```suggestion

def __getattr__(name):
    """Custom getattr to chain and wrap errors due to missing optional deps."""
    raise getattr_missing_optional_dep(
        name, module_name=__name__, optional_names=_missing_optional_names)

```

Moved from the bottom.

qtpy/QtWidgets.py
```diff
@@ -52,3 +79,9 @@
     QApplication.exec_ = QApplication.exec
     QDialog.exec_ = lambda self, *args, **kwargs: self.exec(*args, **kwargs)
     QMenu.exec_ = lambda self, *args, **kwargs: self.exec(*args, **kwargs)
+
+
+def __getattr__(name):
+    """Custom getattr to chain and wrap errors due to missing optional deps."""
+    raise _getattr_missing_optional_dep(
+        name, module_name=__name__, optional_names=_missing_optional_names)
```
ccordoba12, line 87:
```suggestion
```

Moved to the top.

qtpy/utils.py
```diff
@@ -0,0 +1,32 @@
...
+        **wrapper_kwargs,
+    ):
+    """Create a __cause__-chained wrapper error for a missing optional dep."""
+    qtpy_error = wrapper(**wrapper_kwargs)
+    import_error.__cause__ = attr_error
+    qtpy_error.__cause__ = import_error
+    return qtpy_error
+
+
+def _getattr_missing_optional_dep(name, module_name, optional_names):
```
ccordoba12, line 27:
```suggestion
def getattr_missing_optional_dep(name, module_name, optional_names):
```

qtpy/QtGui.py
```diff
@@ -8,7 +8,30 @@
 
 """Provides QtGui classes and functions."""
 
-from . import PYQT6, PYQT5, PYSIDE2, PYSIDE6
+from . import PYQT6, PYQT5, PYSIDE2, PYSIDE6, QtModuleNotInstalledError
+from .utils import _getattr_missing_optional_dep
```
ccordoba12, line 12:
```suggestion
from .utils import getattr_missing_optional_dep
```

I don't think this should be a private function because it's imported and used in other places in our code (like here).

qtpy/QtWidgets.py
```diff
@@ -8,15 +8,31 @@
 
 """Provides widget classes and functions."""
 
-from . import PYQT5, PYQT6, PYSIDE2, PYSIDE6
+from . import PYQT5, PYQT6, PYSIDE2, PYSIDE6, QtModuleNotInstalledError
+from .utils import _getattr_missing_optional_dep
```
ccordoba12, line 12:
```suggestion
from .utils import getattr_missing_optional_dep
```

qtpy/tests/optional_deps/__init__.py
```diff
@@ -0,0 +1,27 @@
+"""Package used for testing the deferred import error mechanism."""
+
+
+# See https://github.com/spyder-ide/qtpy/pull/387/
+
+
+from qtpy.utils import _getattr_missing_optional_dep
```
ccordoba12, line 7:
```suggestion
from qtpy.utils import getattr_missing_optional_dep
```

qtpy/tests/optional_deps/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+    _missing_optional_names['MissingClass'] = {
+        'name': 'optional_dep.MissingClass',
+        'missing_package': 'test_package_please_ignore',
+        'import_error': error,
+    }
+
+
+def __getattr__(name):
+    """Custom getattr to chain and wrap errors due to missing optional deps."""
+    raise _getattr_missing_optional_dep(
```
ccordoba12, line 26:
```suggestion
    raise getattr_missing_optional_dep(
```

ccordoba12 (commented):
Some small suggestions for you @DaelonSuzuka, otherwise looks good to me.

DaelonSuzuka:
> Some small suggestions for you @DaelonSuzuka, otherwise looks good to me.

I like both of those changes, thanks for suggesting them!

ccordoba12 (approved):
Great work here @DaelonSuzuka!

dalthviz (approved):
Thanks for all the work here @DaelonSuzuka !

qtpy/QtGui.py
CAM-Gerlach: Well, it's private to the _package_ as a whole rather than the specific module its defined in. At present, this means it is exposed as a module-level public function in any of the modules its imported in.

PR #391, opened 2022-12-17
PR: Use raise from None when raising QtBindingsNotFoundError in __init__

### LINKED ISSUE #390: QtBindingsNotFoundError is not raised correctly (from None) in __init__
As a followup to issue #367 and PR #379 , I wasn't available to review the latter at the time, but the corrected fix discussed in the former (using `raise QtBindingsNotFoundError from None`)  didn't end up getting implemented, as opposed to the original naive version with several issues. we discussed on the issue that solves the two problems I pointed out with the initial naive implementation was not implemented, making the error/traceback message and behavior more appropriate and less confusing, 

Currently, it is treated and printed as an unexpected error during error handling rather than a intentional wrapper error (with `raise from`), and furthermore prints the error message and traceback from importing `pyside6`, making users think it was incorrectly trying to use that binding instead of the one they selected.

Here's how the error output looks now:

```python
>>> import qtpy.QtCore
  File "C:\Users\C. A. M. Gerlach\Documents\dev\SpyderDev\qtpy\qtpy\__init__.py", line 252, in <module>
ModuleNotFoundError: No module named 'PySide6'

During handling of the above exception, another exception occurred:

  File "C:\Users\C. A. M. Gerlach\Documents\dev\SpyderDev\qtpy\qtpy\__init__.py", line 259, in <module>
    raise QtBindingsNotFoundError()
qtpy.QtBindingsNotFoundError: No Qt bindings could be found
```

and here's how it should look with the correct `raise QtBindingsNotFoundError from None`:

```python
>>> import qtpy.QtCore
  File "C:\Users\C. A. M. Gerlach\Documents\dev\SpyderDev\qtpy\qtpy\__init__.py", line 259, in <module>
    raise QtBindingsNotFoundError from None
qtpy.QtBindingsNotFoundError: No Qt bindings could be found
```

CAM-Gerlach:
As described in #390 , this fixes the `raise QtBindingsNotFoundError` in `__init__,py` to correctly use `raise ... from None`, to ensure it behaves appropriately and has a clear, understandable and non-confusing message.

Here's how the error output looked previously:

```python
>>> import qtpy.QtCore
  File "C:\Users\C. A. M. Gerlach\Documents\dev\SpyderDev\qtpy\qtpy\__init__.py", line 252, in <module>
ModuleNotFoundError: No module named 'PySide6'

During handling of the above exception, another exception occurred:

  File "C:\Users\C. A. M. Gerlach\Documents\dev\SpyderDev\qtpy\qtpy\__init__.py", line 259, in <module>
    raise QtBindingsNotFoundError()
qtpy.QtBindingsNotFoundError: No Qt bindings could be found
```

and here's how it looks now:

```python
>>> import qtpy.QtCore
  File "C:\Users\C. A. M. Gerlach\Documents\dev\SpyderDev\qtpy\qtpy\__init__.py", line 259, in <module>
    raise QtBindingsNotFoundError from None
qtpy.QtBindingsNotFoundError: No Qt bindings could be found
```

Fixes #390

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -256,7 +256,7 @@ if API in PYSIDE6_API:
         QT6 = PYSIDE6 = True
 
     except ImportError:
-        raise QtBindingsNotFoundError()
+        raise QtBindingsNotFoundError from None
     else:
         os.environ[QT_API] = API
 


dalthviz (approved):
Thanks @CAM-Gerlach !

PR #392, opened 2022-12-17
PR: Add Python 3.11 to CIs/tags & overhaul CI config

### LINKED ISSUE #386: Add official support for Python 3.11
Hi,

The recent PySide6 release 6.4.0.1 now supports Python 3.11 too.

Qtpy officially only supports 3.6 to 3.10 but not yet 3.11.

Here's the commit which added 3.10: https://github.com/spyder-ide/qtpy/commit/005641217fbece5e2844ffb844596c90ae3850b6 .

Thanks a lot for this amazing project!

CAM-Gerlach:
As discussed in #386 , adds official support for Python 3.11 (following a similar template to #296 which added 3.10 support) in the Trove tags and CI tests, making the necessary binding and Qt version adjustments as required for the 3.11 jobs, including a somewhat non-ideal but necessary workaround for [a conda package solver incompatibility](https://github.com/spyder-ide/qtpy/pull/392#issuecomment-1356123618) on Py311 + pyqt 5.15 + qt 5.15.

Along the way, I've refactored the CI config to centralize the binding and Qt API version selection all in one place in the matrix config, to greatly reduce duplication of version constants and the scattering of the logic multiple places in the declarative config and the dynamic execution script. In concert with that, I've been able to:

* Ensure that all four bindings have their lower (5.9/6.2), middle (5.12/6.3) and upper (5.15/6.2) bound versions tested on at least one CI job
* Fix an issue with most PySide 6.x jobs accidentally using the lower rather than the middle or upper bound version
* Unskip the macOS-Python3.7-pip job, as it now runs fine without errors
* Add the missing Qt version parameter to the `conda install` call for PySide6 and fix its indent (from PR #378)
* Fix some other issues in `test.sh`, including a small but meaningful typo in a comment, an unnecessary `pip` spec/version, and a tab/spaces problem

I've also updated/fixed a few related small ancillary bits:

* Upgraded the `checkout` and `setup-python` action versions to the current
* Elided the redundant `wheel` build-time dependency
* Removed the deprecated `license_file` metadata key superceded by `license_files`

Fixes #386

### DIFF
--- a/.github/workflows/test.sh
+++ b/.github/workflows/test.sh
@@ -9,8 +9,13 @@ conda remove -q -n test-env --all || true
 # Create and activate conda environment for this test
 BINDING=$(echo "$1" | tr '[:lower:]' '[:upper:]')
 QT_VERSION_VAR=${BINDING}_QT_VERSION
+
 # pytest-qt >=4 doesn't support Qt <=5.9
-if [ "${!QT_VERSION_VAR:0:3}" = "5.9" ]; then PYTESTQT_VERSION="=3.3.0"; PYTEST_VERSION=">=6,!=7.0.0,!=7.0.1,<7.2.0"; fi
+if [ "${!QT_VERSION_VAR:0:3}" = "5.9" ]; then
+    PYTESTQT_VERSION="=3.3.0"
+    PYTEST_VERSION=">=6,!=7.0.0,!=7.0.1,<7.2.0"
+fi
+
 conda create -q -n test-env python=${PYTHON_VERSION} pytest${PYTEST_VERSION:-">=6,!=7.0.0,!=7.0.1"} "pytest-cov>=3.0.0" pytest-qt${PYTESTQT_VERSION:-}
 conda activate test-env
 
--- a/.github/workflows/test.sh
+++ b/.github/workflows/test.sh
@@ -9,19 +9,19 @@ conda remove -q -n test-env --all || true
 # Create and activate conda environment for this test
 BINDING=$(echo "$1" | tr '[:lower:]' '[:upper:]')
 QT_VERSION_VAR=${BINDING}_QT_VERSION
-# pytest-qt >=4 doesn't support Qt >=5.9
+# pytest-qt >=4 doesn't support Qt <=5.9
 if [ "${!QT_VERSION_VAR:0:3}" = "5.9" ]; then PYTESTQT_VERSION="=3.3.0"; PYTEST_VERSION=">=6,!=7.0.0,!=7.0.1,<7.2.0"; fi
-conda create -q -n test-env python=${PYTHON_VERSION} pytest${PYTEST_VERSION:-">=6,!=7.0.0,!=7.0.1"} "pytest-cov>=3.0.0" pytest-qt${PYTESTQT_VERSION:-} pip${PIP_VERSION:-}
+conda create -q -n test-env python=${PYTHON_VERSION} pytest${PYTEST_VERSION:-">=6,!=7.0.0,!=7.0.1"} "pytest-cov>=3.0.0" pytest-qt${PYTESTQT_VERSION:-}
 conda activate test-env
 
 if [ "$USE_CONDA" = "Yes" ]; then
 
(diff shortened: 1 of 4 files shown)


CAM-Gerlach:
@ccordoba12 @dalthviz I've apparently discovered a conda dependency solution issue that can be worked around here, but should probably be fixed properly upstream, most likely with a new release of the `qt` metapackage (which might in turn require a new release of `qt-webengine`). Specifically, no version of the `qt` metapackage and `pyqt` are mutually compatible on Python 3.11, and running

```shell
conda create -n test-env python=3.11 pyqt qt
```

or equivalently

```shell
conda create -n test-env python=3.11 pyqt=5.15 qt=5.15
```

fails on all platforms with a dependency resolution error, which you can see the full output of on CI in e.g. [this run](https://github.com/spyder-ide/qtpy/actions/runs/3718737526/jobs/6307127310).

Specifically, I've traced the issue to following three incompatible constraints:

* The latest `pyqt` 5.15.7 is the earliest version supported on Python 3.11 (5.15.4 will not install)
* `pyqt` 5.15.7 requires `qt-main`, of which only 5.15.6 is mutually compatible with `pyqt` (5.15.4 is mutually incompatible, but will install on its own on Python 3.11, or with `pyqt` 5.15.7on Python <3.11)
* The latest version of the `qt` 5.15 meta-package is 5.15.4, which pins on `qt-main` and `qt-webengine` of the same point version

To note,

```shell
conda create -n test-env python=3.11 pyqt=5.15.7 qt-main=5.15.6 qt-webengine=5.15.4
```

does resolve. Also, 5.15.4 is the latest version of `qt-webengine`, though conda-forge/qt-webengine-feedstock#20 is open to update it to 5.15.7—though that wouldn't be in sync with the latest `qt-main` version, so I'm not sure if that can be pinned with `qt` unless `qt-main` is also updated to that version.

Should this be reported to one or more of those feedstocks? Or can something else be done about it?

ccordoba12:
> Should this be reported to one or more of those feedstocks? Or can something else be done about it?

I think it should be reported in the qt-feedstock to update the constraints it has on `qt-main` and `qt-webengine`.

CAM-Gerlach:
Thanks, I've opened conda-forge/qt-feedstock#234 on there as suggested.

In the meantime, though, that doesn't block this PR, since I implemented a workaround (that's not quite as elegant as I'd like, but works just fine for now).

.github/workflows/test.sh
```diff
@@ -9,36 +9,41 @@ conda remove -q -n test-env --all || true
 # Create and activate conda environment for this test
 BINDING=$(echo "$1" | tr '[:lower:]' '[:upper:]')
 QT_VERSION_VAR=${BINDING}_QT_VERSION
-# pytest-qt >=4 doesn't support Qt >=5.9
+# pytest-qt >=4 doesn't support Qt <=5.9
 if [ "${!QT_VERSION_VAR:0:3}" = "5.9" ]; then PYTESTQT_VERSION="=3.3.0"; PYTEST_VERSION=">=6,!=7.0.0,!=7.0.1,<7.2.0"; fi
```
ccordoba12, line 13:
```suggestion

# pytest-qt >=4 doesn't support Qt <=5.9
if [ "${!QT_VERSION_VAR:0:3}" = "5.9" ]; then
    PYTESTQT_VERSION="=3.3.0"
    PYTEST_VERSION=">=6,!=7.0.0,!=7.0.1,<7.2.0"
fi

```

No need for this `if` to be in a single line, which makes it hard to read.

ccordoba12 (commented):
A couple of small suggestions for you @CAM-Gerlach, otherwise looks good to me.

.github/workflows/test.sh
CAM-Gerlach: Yeah, I'm not sure why it was that way originally; it was added at some point after I created the original script

ccordoba12 (approved):
Looks good to me now, thanks @CAM-Gerlach!

dalthviz (approved):
Thanks @CAM-Gerlach !

PR #393, opened 2022-12-17
PR: Fix `LibraryLocation` -> `LibraryPath` renaming due to deprecation with Qt6

StSav012:
That was a minor omission.

### DIFF
--- a/qtpy/QtCore.py
+++ b/qtpy/QtCore.py
@@ -63,6 +63,7 @@ elif PYQT6:
     QTextStreamManipulator.exec_ = lambda self, *args, **kwargs: self.exec(*args, **kwargs)
 
     QLibraryInfo.location = QLibraryInfo.path
+    QLibraryInfo.LibraryLocation = QLibraryInfo.LibraryPath
 
     # Those are imported from `import *`
     del pyqtSignal, pyqtBoundSignal, pyqtSlot, pyqtProperty, QT_VERSION_STR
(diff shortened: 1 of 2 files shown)


StSav012:
More info on the change: https://doc.qt.io/qt-6/qlibraryinfo-obsolete.html

dalthviz (commented):
Hi @StSav012 thank you for the help with this! I think we should add a test for this. It should go inside [`test_qtcore.py`](https://github.com/spyder-ide/qtpy/blob/master/qtpy/tests/test_qtcore.py) and be something similar to the tests we have for `QLibraryInfo.location`: https://github.com/spyder-ide/qtpy/blob/371ec52e6b687ae89b3e422ff2a13cba28d4bd85/qtpy/tests/test_qtcore.py#L65-L68

Other than that, this LGTM :+1: If you have any question to implement the test let us know!

StSav012:
Hi @dalthviz,

Thank you for the feedback. Forgive me for not looking through the tests, for I'm not disciplines enough (yet) to make tests for my code.

As a side note, is line https://github.com/spyder-ide/qtpy/blob/371ec52e6b687ae89b3e422ff2a13cba28d4bd85/qtpy/tests/test_qtcore.py#L73 in the tests correct? I would expect an `assert` there, like you have in similar places. On the other hand, if there is no `exec_` function, the expression should raise an `AttributeError`, right? Then, many `assert`s aren't necessary at all.

dalthviz:
> in the tests correct? I would expect an assert there, like you have in similar places. On the other hand, if there is no exec_ function, the expression should raise an AttributeError, right? Then, many asserts aren't necessary at all.

You are probably right there @StSav012 ! In general I believe we have a lot of room for improvement regarding testing 😅 If you want to help with that or have any ideas feel free to open a new issue to discuss and maybe plan what changes we could do to improve our test suite :)

dalthviz (approved):
Thanks @StSav012 !

PR #397, opened 2023-01-12
Simplify CI script following Qt 5.15.6 feedstock update

CAM-Gerlach:
Followup to #392 

As conda-forge/qt-feedstock#234 seems to be resolved by way of conda-forge/qt-feedstock#235 , we can now remove the workaround in our CI needed for that issue, which re-simplifies the script quite a bit.

### DIFF
--- a/.github/workflows/test.sh
+++ b/.github/workflows/test.sh
@@ -22,12 +22,7 @@ conda activate test-env
 if [ "$USE_CONDA" = "Yes" ]; then
 
     if [ "${1}" = "pyqt5" ]; then
-        if [ "${PYQT5_VERSION:0:4}" = "5.15" ]; then
-            # Must specify versions individually with PyQt 5.15 to avoid compat issues with Python 3.11
-            conda install -q qt-main=${PYQT5_QT_VERSION} qt-webengine=${PYQT5_QT_VERSION} pyqt=${PYQT5_VERSION}
-        else
-            conda install -q qt=${PYQT5_QT_VERSION} pyqt=${PYQT5_VERSION}
-        fi
+        conda install -q qt=${PYQT5_QT_VERSION} pyqt=${PYQT5_VERSION}
     elif [ "${1}" = "pyside2" ]; then
         conda install -q qt=${PYSIDE2_QT_VERSION} pyside2=${PYSIDE2_VERSION}
     elif [ "${1}" = "pyside6" ]; then


ccordoba12 (approved):

.github/workflows/test.sh
```diff
@@ -22,12 +22,7 @@ conda activate test-env
 if [ "$USE_CONDA" = "Yes" ]; then
 
     if [ "${1}" = "pyqt5" ]; then
-        if [ "${PYQT5_VERSION:0:4}" = "5.15" ]; then
-            # Must specify versions individually with PyQt 5.15 to avoid compat issues with Python 3.11
-            conda install -q qt-main=${PYQT5_QT_VERSION} qt-webengine=${PYQT5_QT_VERSION} pyqt=${PYQT5_VERSION}
-        else
-            conda install -q qt=${PYQT5_QT_VERSION} pyqt=${PYQT5_VERSION}
-        fi
+        conda install -q qt=${PYQT5_QT_VERSION} pyqt=${PYQT5_VERSION}
```
hmaarrfk, line 25:
```suggestion
        conda install -q qt-main=${PYQT5_QT_VERSION} qt-webengine=5.15 pyqt=${PYQT5_VERSION}
```
Why not do this?

This is what we do in the qt package. Basically, i couldn't get the version numbers of qt-webengine and qt-main to match up exactly.

.github/workflows/test.sh
```diff
@@ -22,12 +22,7 @@ conda activate test-env
...
     elif [ "${1}" = "pyside2" ]; then
         conda install -q qt=${PYSIDE2_QT_VERSION} pyside2=${PYSIDE2_VERSION}
```
hmaarrfk, line 27:
```suggestion
        conda install -q qt-main=${PYSIDE2_QT_VERSION} qt-webengine=5.15 pyside2=${PYSIDE2_VERSION}
```

hmaarrfk:
I'm trying to keep up with qt (and tried to get some help from the bot more recently) but there are just more eyes on the qt-main and qt-webengine feedstocks.

.github/workflows/test.sh
CAM-Gerlach: Because `qt-main` is only available for Qt 5.15.x, and not anything else that we're testing in our matrix (5.9, 5.12, etc). If we do this, we need to have separate logic branches for each version, which is exactly what the script does now, and this PR eliminates.

The goal is to avoid as much fragile, complex, duplicative and non-trivial to maintain version-specific branching logic in this script as possible, and instead keep all of that in the build matrix in the declarative workflow config file (there are still a few bits that should be factored out, but that's a subject for another PR).

Also, I'm not sure why `qt-webengine` is inexplicably hardcoded, instead of using the existing variable, i.e. just using the original line here.

.github/workflows/test.sh
CAM-Gerlach: See above—because `qt-main` is only available for Qt 5.15.x, so we'd have to add the same fragile, complex, duplicative and tedious to maintain branching logic that we had above to here as well if we were to do this.

CAM-Gerlach: To note, that is indeed exactly what we do for PySide6, and will presumably be the case for PyQt6, as being a whole separate matrix job, package, and variables, the logic branch is much more justifiable and required anyway.

.github/workflows/test.sh
hmaarrfk: Understood

dalthviz (approved):

Also, thank you @hmaarrfk for the review! If you have any other suggestions or questions you want to discuss let us know!

CAM-Gerlach:
> I'm trying to keep up with qt (and tried to get some help from the bot more recently) but there are just more eyes on the qt-main and qt-webengine feedstocks.

And by the way, thanks so much for your help with all that @hmaarrfk !

Issue #395, opened 2023-01-02
Binding error compatabilty issues


jain-brain:
I am receiving the following error. I have searched the internet everywhere for the solution. Can you please tell me how to resolve this? Is this due to incompatibility of versions of the library?

`/lib/python3.9/site-packages/qtpy/__init__.py:267: RuntimeWarning: Selected binding "pyside6" could not be found, using "pyqt5"
  warnings.warn('Selected binding "{}" could not be found, '
`

CAM-Gerlach:
From the minimal information provided above, this itself is just a warning, rather than an error, indicating that QtPy could not find an installed Qt binding (PyQt or PySide) in your current environment. The simple solution is to simply install one (PyQt5 or 6, or PySide2 or 6) using whatever package manager (pip, conda or OS) you are already using for your Python environment (virtualenv, conda env, etc).

Beyond that, we'll need a lot more information to be able to provide more specific advice. In particular:

* What are you trying to do?
* What application are you trying to use QtPy with?
* In what circumstances does the warning appear?
* What other behavior/errors are you seeing?
* What operating system version and versions of Python, Qt, PyQt/PySide, and QtPy are you using?
* How did you install Python (Python.org, WinPython, Anaconda, distro, etc) and QtPy (Anaconda defaults? Conda-Forge? pip (conda env)? pip (virtualenv)? GitHub clone? Distro package manager?)
 
Also, please paste the full commands and output you are attempting inside a code block, i.e.

~~~markdown
```
<Paste full commands and output here>
```
~~~

And also paste the full `conda list` (if using a `conda` based Python distribution) or `pip list` (otherwise) output in the same:

~~~markdown
```
<Paste full pip/conda list output>
```
~~~

Thanks!

dalthviz:
Could it be worthy to add some descriptions to the README regarding possible warnings that QtPy can show @CAM-Gerlach ?

CAM-Gerlach:
It would probably be worth fully documenting the error and warning hierarchy defined in `__init__`, yeah, at least at some point. However, at least in terms of end user interpretation (as appears to be the case here), the messages themselves should be clear and easy enough to understand on their own first and foremost—and right now, there's only one general warning defined, with a rather inaccurate/outdated docstring, and it isn't even used here (instead, a generic `RuntimeWarning` is used).

To clean up the warning class, message and docstring while remaining fully backwards compatible, I've opened PR #398

dalthviz:
[...]

PR #398, opened 2023-01-14
PR: Make warning usage consistant and refine messages

CAM-Gerlach:
Currently, as noted in issue #395 , a generic `RuntimeWarning` is emitted when falling back to a non-selected binding, instead of QtPy's own `PythonQtWarning` subclass. Furthermore, the text could perhaps be clarified a bit. I also noticed a few other issues when making this changed, which I fixed.

Therefore, this PR:

* Emits a `PythonQtWarning` instead of a generic `RuntimeWarning` when falling back to an unselected binding, consistent with the other warnings and errors that use QtPy-specific subclasses
* Has `PythonQtWarning` inherit from `RuntimeWarning` instead of a generic `Warning` , to be consistent with all the other errors and warnings and with what `RuntimeWarning` is meant for, and ensure the previous replacement is fully backwards compatible
* Tweaks the message text of the former warning to be clearer and more grammatically correct
* Fixes the outdated and incorrect docstrings of `PythonQtWarning` and `PythonQtError`
* Updates both warning usages to modern code formatting, f-strings and to use `!r` instead of hardcoded quotes

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -65,11 +65,11 @@ __version__ = '2.4.0.dev0'
 
 
 class PythonQtError(RuntimeError):
-    """Error raised if no bindings could be selected."""
+    """Generic error superclass for QtPy."""
 
 
-class PythonQtWarning(Warning):
-    """Warning if some features are not implemented in a binding."""
+class PythonQtWarning(RuntimeWarning):
+    """Warning class for QtPy."""
 
 
 class PythonQtValueError(ValueError):
@@ -79,7 +79,7 @@ class PythonQtValueError(ValueError):
 class QtBindingsNotFoundError(PythonQtError):
     """Error raised if no bindings could be selected."""
     _msg = 'No Qt bindings could be found'
-    
+
     def __init__(self):
         super().__init__(self._msg)
 
@@ -106,7 +106,7 @@ class QtModuleNotInOSError(QtModuleNotFoundError):
 class QtModuleNotInQtVersionError(QtModuleNotFoundError):
     """Raised when a module is not implemented in the current Qt version."""
     _msg = '{name} does not exist in {version}.'
-    
+
     def __init__(self, *, name, msg=None, **msg_kwargs):
         global QT5, QT6
         version = 'Qt5' if QT5 else 'Qt6'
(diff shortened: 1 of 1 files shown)


CAM-Gerlach:
(Oops, fell asleep before I went back and hit the ready for review button)

qtpy/__init__.py
```diff
@@ -264,8 +264,11 @@ def __init__(self, *, missing_package=None, **superclass_kwargs):
 # If a correct API name is passed to QT_API and it could not be found,
 # switches to another and informs through the warning
 if API != initial_api and binding_specified:
-    warnings.warn('Selected binding "{}" could not be found, '
-                  'using "{}"'.format(initial_api, API), RuntimeWarning)
+    warnings.warn(
+        f'Selected binding {initial_api!r} could not be found; '
+        f'falling back to {API!r}',
+        PythonQtWarning,
+        )
```
ccordoba12, line 271:
```suggestion
    )
```

qtpy/__init__.py
```diff
@@ -282,10 +285,10 @@ def __init__(self, *, missing_package=None, **superclass_kwargs):
...
     """Warn if using a Qt or binding version no longer supported by QtPy."""
     warning_message = (
-        "{name} version {old_version} is not supported by QtPy. "
-        "To ensure your application works correctly with QtPy, "
-        "please upgrade to {name} {min_version} or later.".format(
-            name=name, old_version=old_version, min_version=min_version))
+        f'{name} version {old_version} is not supported by QtPy. '
+        'To ensure your application works correctly with QtPy, '
+        f'please upgrade to {name} {min_version} or later.'
+        )
```
ccordoba12, line 291:
```suggestion
    )
```

ccordoba12 (commented):
Two small suggestions for you @CAM-Gerlach, otherwise looks good to me.

qtpy/__init__.py
CAM-Gerlach: Rebased in; I was just using the Spyder default style rather than this Black style since we were going to run Black anyway and I thought other code here used it, but sure :+1:

ccordoba12 (approved):
Looks good to me now, thanks @CAM-Gerlach!

dalthviz (approved):
Thanks @CAM-Gerlach !

dalthviz:
Also @CAM-Gerlach should this close #395 ?

CAM-Gerlach:
It may help with it some, but it isn't clear that it would resolve the actual issue the user experienced, since from the very limited information provided it is not at all clear what that actually is. I suggest we wait at least a couple weeks to see if they respond. Also, you mentioned documenting the error/warning hierarchy somewhere, which is a good idea to do at some point.

PR #400, opened 2023-02-01
Remove PySide2 from APIs on Python 3.11+

StSav012:
As far as I can judge from the [`PySide*` chat message](https://t.me/qtforpython/19724) by Florian Bruhin, PySide2 is not going to run on Python 3.11+. Currently, it crashes Python. To avoid further issues, I suggest removing it from the APIs for Python 3.11 and up.

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -157,6 +157,8 @@ binding_specified = QT_API in os.environ
 
 API_NAMES = {'pyqt5': 'PyQt5', 'pyside2': 'PySide2',
              'pyqt6': 'PyQt6', 'pyside6': 'PySide6'}
+if sys.version_info >= (3, 11):
+    API_NAMES.pop('pyside2')  # PySide2 does not support Python 3.11 and newer
 API = os.environ.get(QT_API, 'pyqt5').lower()
 initial_api = API
 if API not in API_NAMES:
--- a/qtpy/tests/test_cli.py
+++ b/qtpy/tests/test_cli.py
@@ -44,34 +44,37 @@ def test_cli_mypy_args():
     )
 
     if qtpy.PYQT5:
-        expected = ' '.join([
+        expected = [
             '--always-true=PYQT5',
             '--always-false=PYSIDE2',
             '--always-false=PYQT6',
             '--always-false=PYSIDE6',
-        ])
+        ]
     elif qtpy.PYSIDE2:
-        expected = ' '.join([
+        expected = [
             '--always-false=PYQT5',
             '--always-true=PYSIDE2',
             '--always-false=PYQT6',
             '--always-false=PYSIDE6',
-        ])
+        ]
     elif qtpy.PYQT6:
-        expected = ' '.join([
+        expected = [
             '--always-false=PYQT5',
             '--always-false=PYSIDE2',
             '--always-true=PYQT6',
             '--always-false=PYSIDE6',
-        ])
+        ]
     elif qtpy.PYSIDE6:
-        expected = ' '.join([
+        expected = [
             '--always-false=PYQT5',
             '--always-false=PYSIDE2',
             '--always-false=PYQT6',
             '--always-true=PYSIDE6',
-        ])
+        ]
     else:
         assert False, 'No valid API to test'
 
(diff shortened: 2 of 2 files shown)


dalthviz:
Hi @StSav012 thank you checking this! Although I'm a little bit confused, it is possible to install PySide2 with pip on Python 3.11? I think you can't since there are no compatible wheels available, right?

image

Also, seems like you can install PySide2 5.15.8 with conda (conda-forge) and after testing locally seems like it works with Python 3.11 🤔  

image

Also, just in case, what do you think about this @CAM-Gerlach @ccordoba12 ?

StSav012:
I'm actually surprised. First, in the tests, there is Python 3.10.8 (see conda info) despite the tests are titled Python 3.11. Then, by the fact that the developers are unaware of `PySide2` coming with `conda`. I'd check `sys.version_info` in the `conda` environment.

As a side note, how old Python, PySide2, and PyQt5 do you support? In the code, I see some lines about PyQt5 5.9 and 5.12, and I can PR patches for PyQt5 5.6 if someone needs them.

dalthviz:
> I'm actually surprised. First, in the tests, there is Python 3.10.8 (see conda info) despite the tests are titled Python 3.11.

You are totally right! The CI is not installing Python 3.11 in the test env, we need to fix that :+1:

> As a side note, how old Python, PySide2, and PyQt5 do you support?

PySide2 5.12 and PyQt5 5.9. For the validations those version are defined here: https://github.com/spyder-ide/qtpy/blob/aa3fe4bff4d629f93b2e74bbf6550a6793150431/qtpy/__init__.py#L146-L149

> I can use PySide2 on Python 3.11 almost right, except for a crash when an app quits. So, for the end user, there is a mere difference.

Oh then I think we should create an issue in the PySide2 feedstock repo: https://github.com/conda-forge/pyside2-feedstock

ccordoba12:
> Also, just in case, what do you think about this @CAM-Gerlach @ccordoba12 ?

If PySide2 is working fine with Python 3.11 with Conda-forge packages, then we certainly don't need to do this.

> Then, by the fact that the developers are unaware of PySide2 coming with conda

That's no surprise for us because there's an important disconnect between developers outside of the Scientific Python ecosystem and those inside it.

CAM-Gerlach:
Right now, we actually aren't testing PySide2 on Python 3.11 on Conda, as I hadn't gotten it to work in PR #392 , but [I tested it just now](https://github.com/CAM-Gerlach/qtpy/actions/runs/4070567890/jobs/7011545685) and it works just fine, so I suspect that was a result of #397 . I've opened PR #401 to unskip it.

In addition, if we did want to do it, this wouldn't be the way to go about it. This PR removes it from the lookup table of API names, which ensures that a somewhat ambiguous error message will be immediately presented if the user manually specifies it with the `QT_API` environment variable, and the MyPy CLI output will not include it (as it should, given the global variable is still defined and used), but otherwise it will still be searched for and attempted to be used regardless, potentially triggering the hard crash, rather than falling back to another usable API, which would be the most potentially useful part of this.

> Although I'm a little bit confused, it is possible to install PySide2 with pip on Python 3.11? I think you can't since there are no compatible wheels available, right?

Yeah, there are no wheels past 3.10, and no sidsts at all, so the only way a user would install it would be either through conda (which works) or through manually building their own source checkout (which they could patch to fix). Therefore, it's not clear to me there's any user benefit to restricting it.

> I think the Python 3.10.8 info you see on the CI is from the initial conda installation that is created when setting up conda with the `setup-miniconda` action (which runs before the [`test.sh` script](https://github.com/spyder-ide/qtpy/blob/master/.github/workflows/test.sh?rgh-link-date=2023-02-01T21%3A10%3A19Z))

Yup, that's right.

StSav012:
Your replies make perfect sense. Thank you. I drop the PR.

dalthviz:
Thank you @StSav012 for raising up this possible source of issues and thinking about a possible solution! If you have further concerns in the future about possible things rasing issues let us know!

CAM-Gerlach:
And it was also thanks to this PR that we opened #401 to unskip PySide2 Py3.11 testing in CI on Conda, as it is now working but was previously skipped.

PR #401, opened 2023-02-02
PR: Unskip PySide2 5.15 on Python 3.11 with Conda on CIs

CAM-Gerlach:
As discussed on #400 and a followup to PR #392, unskip PySide2 on Python 3.11, because it works just fine now (at least after #397).


ccordoba12 (approved):
Thanks @CAM-Gerlach!

dalthviz (approved):
Thanks @CAM-Gerlach

PR #403, opened 2023-02-02
PR: Backport Qt6 behavior to PyQt5 and PySide2

StSav012:
Changes:

- fixed enums in `PyQt5==5.9.3`; there had been tests skipped because of them;
- fixed types for `setTabStopDistance` and `tabStopDistance` of `QTextEdit` and `QPlainTextEdit`;
- fixed results of `QLocale.toLong` and `QLocale.toULong`;
- fixed `QMenu.exec` when called as a static function;
- fixed `QWizardPage.registerField` to accept a string as `changedSignal` argument;
- placed `QAction`, `QActionGroup`, `QFileSystemModel`, `QShortcut`, and `QUndoCommand` both into `QtGui` and `QtWidgets`;
- placed `QFileSystemModel` into `QtGui`;
- added tests for `exec`/`print` as well as `exec_`/`print_` methods;
- added `toPyTimeDate`/`toPyDate`/`toPyTime`  alongside `toPython` to unify the code coming from different Qt flavors;
- added `QtCore.qEnvironmentVariable`;
- added `QByteArray.chopped`, `QByteArray.isUpper`, `QByteArray.isLower`, `QByteArray.Base64Option`, `QByteArray.Base64DecodingStatus`, and `QByteArray.fromBase64Encoding`;
- added `QDate.startOfDay` and `QDate.endOfDay`;
- added `QDateTime.YearRange` (not good enough, though);
- added `QLibraryInfo.path` alongside `QLibraryInfo.location`;
- added `QLine.toLineF`, `QMargins.toMarginsF`, `QPoint.toPointF`, `QRect.toRectF`, and `QSize.toSizeF` functions;
- added `QSize.grownBy`, `QSize.shrunkBy`, `QSizeF.grownBy`, and `QSizeF.shrunkBy`;
- added `QModelIndex.siblingAtColumn` and `QModelIndex.siblingAtRow` functions;
- added `QRegularExpression.anchoredPattern` and `QRegularExpression.wildcardToRegularExpression`;
- added `QColor.toTuple`, `QColor.fromString`, and `QColor.isValidColorName` functions;
- added `QImage.sizeInBytes`;
- added `QMouseEvent.position` from `QMouseEvent.pos`;
- added `QWheelEvent.position` and `QWheelEvent.globalPosition` (TODO: unify the tests with ones for `QMouseEvent`);
- added `QTabletEvent.deviceType`;
- added `QFontMetrics.horizontalAdvance` and `QFontMetricsF.horizontalAdvance`;
- adding `QtGui.QColorConstants` (WIP);
- removed tests for `QPlainTextEdit.tabStopWidth` and `QTextEdit.tabStopWidth` on PyQt5<5.10.

I do feel it's far too much for a single PR.

P. S. I have no idea why the tests report failures, whereas my local ones don't.


CAM-Gerlach:
> P. S. I have no idea why the tests report failures, whereas my local ones don't.

Seems like the crash occurs on the new `menu.exec()` line of the `test_qmenu_functions` test on PySide2 for all the platforms in which it is currently unskipped (Python 3.7 on Linux with Conda and on Windows with and without Conda). We've been able to successfully unskip the PySide2 job for Python 3.11 on Conda; could you rebase this PR on ``master`` (i.e. `git fetch upstream && git rebase upstream/master && git push -f`) to re-run with those jobs enabled, so that we can see if it also fails on Python 3.11?

Not sure exactly what is going on, but you could maybe try using the `lambda` approach instead of direct re-assignment?

qtpy/QtCore.py
```diff
@@ -20,14 +20,27 @@
...
     QDateTime.toPython = lambda self, *args, **kwargs: self.toPyDateTime(*args, **kwargs)
     QTime.toPython = lambda self, *args, **kwargs: self.toPyTime(*args, **kwargs)
 
+    # Fix enums in PyQt5 5.9.*
+    from .enums_compat import demote_enums
+    if PYQT5 and __version__.startswith('5.9.'):
+        from PyQt5 import QtCore
+        demote_enums(QtCore)
+        del QtCore
+    del demote_enums
```
CAM-Gerlach, line 33:
```suggestion
    if __version__.startswith('5.9.'):
        from .enums_compat import demote_enums
        from PyQt5 import QtCore
        demote_enums(QtCore)
        del QtCore, demote_enums
```

* Move `.enums_compt` import and deletion inside the block, so it's only executed if nessesary
* Remove redundant `PYQT5` check (since we're already inside the `PYQT5` block)

qtpy/QtGui.py
```diff
@@ -12,6 +12,28 @@
...
+        QFontMetricsF.horizontalAdvance = lambda self, *args, **kwargs: self.width(*args, **kwargs)
+
+    # Fix enums in PyQt5 5.9.*
+    from .enums_compat import demote_enums
+    from PyQt5.QtCore import QT_VERSION_STR as __version__
+    if PYQT5 and __version__.startswith('5.9.'):
+        from PyQt5 import QtGui
+        demote_enums(QtGui)
+        del QtGui
+    del __version__, demote_enums
```
CAM-Gerlach, line 35:
```suggestion
    from PyQt5.QtCore import QT_VERSION_STR as __version__
    if __version__.startswith('5.9.'):
        from .enums_compat import demote_enums
        from PyQt5 import QtGui
        demote_enums(QtGui)
        del QtGui, demote_enums
    del __version__
```

* Move `.enums_compt` import and deletion inside the block, so it's only executed if necessary
* Remove redundant `PYQT5` check (since we're already inside the `PYQT5` block)

qtpy/QtWidgets.py
```diff
@@ -12,18 +12,27 @@
...
     from PyQt5.QtWidgets import *
+
+    # Fix enums in PyQt5 5.9.*
+    from .enums_compat import demote_enums
+    from PyQt5.QtCore import QT_VERSION_STR as __version__
+    if PYQT5 and __version__.startswith('5.9.'):
+        from PyQt5 import QtWidgets
+        demote_enums(QtWidgets)
+        del QtWidgets
+    del __version__, demote_enums
```
CAM-Gerlach, line 23:
```suggestion
    from PyQt5.QtCore import QT_VERSION_STR as __version__
    if __version__.startswith('5.9.'):
        from .enums_compat import demote_enums
        from PyQt5 import QtWidgets
        demote_enums(QtWidgets)
        del QtWidgets, demote_enums
    del __version__
```

* Move `.enums_compt` import and deletion inside the block, so it's only executed if necessary
* Remove redundant `PYQT5` check (since we're already inside the `PYQT5` block)

qtpy/enums_compat.py
```diff
@@ -37,3 +37,24 @@ def promote_enums(module):
...
+        Search unscoped enums in the given module and allow scoped access.
+
+        Special for PyQt5==5.9.*
+        """
+        class LookUp:
+            def __init__(self, lookup_location):
+                self._ll = lookup_location
+
+            def __getattr__(self, what):
+                return getattr(self._ll, what)
```
CAM-Gerlach, line 53:
```suggestion
                self._lookup_location = lookup_location

            def __getattr__(self, what):
                return getattr(self._lookup_location, what)
```

Any reason not to use a more descriptive name here?

qtpy/enums_compat.py
```diff
@@ -37,3 +37,24 @@ def promote_enums(module):
...
+
+        for class_name in dir(module):
+            if class_name == 'Qt' or (class_name[0] == 'Q' and class_name[1].isupper()):
+                klass = getattr(module, class_name)
+                for attrib_name in dir(klass):
+                    if attrib_name[0].isupper() and type(getattr(klass, attrib_name)).__name__ == 'enum''type':
```
CAM-Gerlach, line 59:
```suggestion
                    if attrib_name[0].isupper() and type(getattr(klass, attrib_name)).__name__ == 'enumtype':
```

Looks like maybe an issue with Black's older string parsing?

qtpy/tests/test_qtcore.py
```diff
@@ -46,36 +46,82 @@ def test_qtime_toPython():
     assert isinstance(py_time, time)
 
 
+def test_qdatetime_toPyDateTime():
+    """Test QDateTime.toPyDateTime"""
+    q_date = QtCore.QDateTime.currentDateTime()
+    assert QtCore.QDateTime.toPyDateTime is not None
+    py_date = q_date.toPyDateTime()
+    assert isinstance(py_date, datetime)
```
CAM-Gerlach, line 54: Maybe worth passing a fixed datetime (and similar, for the others) and checking if the resulting object equals the expected?

CAM-Gerlach (commented):
**Standard reminder: You can directly apply all the suggestions you want in one go with [`Files changed`](https://github.com/spyder-ide/qtpy/pull/403/files) -> <kbd>Add to batch</kbd> -> <kbd>Commit</kbd>**

Some relatively minor comments, as GitHub suggestions. @dalthviz or @ccordoba12 might have more substantive feedback.

qtpy/enums_compat.py
StSav012: I've split the string just to avoid a typo reported by the IDE.

I know that the part looks ugly, but I can't think of another way of checking the type against `sip.enumtype`.

qtpy/enums_compat.py
StSav012: The “l”s stand for my laziness. That's the only reason.

Actually, I wrote the code on Windows XP for `PyQt5==5.6` (that was painful). There was next to no IDE, so I used as short variable names as were possible, keeping the code understandable.

qtpy/tests/test_qtcore.py
StSav012: Makes sense. This way, we'll check that the conversion is valid.

StSav012:
> Not sure exactly what is going on, but you could maybe try using the `lambda` approach instead of direct re-assignment?

When I assign `exec` of `QApplication` directly, I can make a `QApplication` with a `QMainWindow` manually just fine. It doesn't look like an issue. However, when I `exec` (not `exec_`) a menu (with or without a point specified), I catch a segfault, regardless of whether I do `app.exec()` by the time or not. The `lambda` resolves the issue. So, I'll convert all the assignments to lambdas just in case.

StSav012:
After changing to a `lambda`, I face the following error on PySide2 on Python 3.11 only:

```plaintext
>   QMenu.exec = lambda self, *args, **kwargs: self.exec_(*args, **kwargs)
E   ValueError: 'PySide2.QtWidgets.QMenu.exec_' called with wrong argument values:
E     PySide2.QtWidgets.QMenu.exec_()
E   Found signature:
E     PySide2.QtWidgets.QMenu.exec_()

/usr/share/miniconda/envs/test-env/lib/python3.11/site-packages/qtpy/QtWidgets.py:52: ValueError
```

<details><summary>They are the same!</summary>
<p>

[image](https://knowyourmeme.com/memes/theyre-the-same-picture)

</p>
</details>

It turns out, that `QMenu.exec_` can actually be static:
```python
@staticmethod
def exec_(actions: typing.Sequence[PySide2.QtWidgets.QAction], pos: PySide2.QtCore.QPoint, at: typing.Optional[PySide2.QtWidgets.QAction] = None, parent: typing.Optional[PySide2.QtWidgets.QWidget] = None) -> PySide2.QtWidgets.QAction:
    pass
```
Might it be the reason? It seems, not, although I've fixed the omission. I'm at a loss.

qtpy/enums_compat.py
CAM-Gerlach: Windows XP :scream:

Oh wow, how long ago did you write this code?

qtpy/enums_compat.py
CAM-Gerlach: Could you silence that with a `# noqa` comment or similar? Because otherwise it looks like a typo from my *eye*-DEs (:laughing: ). Not sure its worth this ugly hack just because one particular IDE has a false positive...

CAM-Gerlach:
Hmm, that's way out of my depth...there doesn't appear to be an obvious solution other than those you've tried, at least without digging through the PySide2 code to see what's going on. @dalthviz , any ideas?

If all else fails, we could skip the test on Python 3.11 for now and open an issue to come back to it later...though with all the work you've put into trying to fix it, I would hate to have to do that.

qtpy/enums_compat.py
StSav012: Some 2.5 _months_ ago. I had to port lots of code from Python 3.11 with PySide6 back to Python 3.4 and PyQt5 5.6. For that, I had to make up a tool to speed it up. The tool worked so well that I didn't have to fix anything manually most of the time. That was so great and useless project!

CAM-Gerlach: Wow, I can't imagine what sort of project would require that...considering it's been EoL for nearly a decade now (aside from the last embedded/POS editions, which "only" went EOL 4 years ago). In any case, doesn't sound like fun, though seems like your tool managed to make it less of a pain.

StSav012:
@CAM-Gerlach, I'm close to nailing it down! Consider the following code:

```python
from PySide2.QtCore import QPoint, QTimer
from PySide2.QtWidgets import QApplication, QMenu


def possibly_static_exec_(cls, *args, **kwargs):
    """ Call `self.exec_` when `self` is given or a static method otherwise. """
    if isinstance(args[0], cls):
        return args[0].exec_(*args[1:], **kwargs)
    else:
        return cls.exec_(*args, **kwargs)


QMenu.exec = lambda *args, **kwargs: possibly_static_exec_(QMenu, *args, **kwargs)
app = QApplication([])
m = QMenu(None)
m.addAction('Action')
QTimer.singleShot(1000, m.close)
m.exec(QPoint(1, 1))
```

It runs fine, with no errors and the exit code == 0. Notice `m.exec(QPoint(1, 1))` there. If there is `m.exec()` instead, I face the same `ValueError`. Finally, I get it locally.

Then I patched the `possibly_static_exec_` function for that very case:
```python
def possibly_static_exec_(cls, *args, **kwargs):
    """ Call `self.exec_` when `self` is given or a static method otherwise. """
    if isinstance(args[0], cls):
        if len(args) == 1 and not kwargs:
            return args[0].exec_()
        return args[0].exec_(*args[1:], **kwargs)
    else:
        return cls.exec_(*args, **kwargs)
```
It runs without an error! 

Still, I'm uncertain whether the code runs in all possible cases. It would be wonderful if you compose tests for the function.

dalthviz (commented):
Thank you so much for all the work here @StSav012 ! 

Regarding the way to test the static version of the `QMenu.exec` maybe we could use something like:

```python
from PyQt5.QtWidgets import QMenu, QAction, QApplication, QWidget
from PyQt5.QtCore import QPoint, QTimer, Qt
from PyQt5.QtGui import QMouseEvent
from PyQt5.QtTest import QTest
qapp = QApplication([])

action = QAction("text")
QTimer.singleShot(1000, lambda: QTest.keyClick(
    QApplication.widgetAt(1,1),
    Qt.Key.Key_Escape)
)
QMenu.exec_([action], QPoint(1, 1))
```

But using the `qtbot` fixture instead of `QTest` ?

And we could use something like https://pytest-mock.readthedocs.io/en/latest to be sure in an easy way that `possibly_static_exec_` is being call?

Also, could you fix the merge conflict please?

StSav012:
> Also, could you fix the merge conflict please?

My bad. I should have been more patient to avoid the conflict in the first place.

> And we could use something like https://pytest-mock.readthedocs.io/en/latest to be sure in an easy way that possibly_static_exec_ is being call?

Do we actually need that? If both `menu.exec()` and `QMenu.exec(...)` work, does it matter, why?

StSav012:
I've marked the PR as a draft. To my horror, I've faced the same gremlin of `QMenu` again today. The following run of the tests has gone fine in that aspect.

As the PR is planned only for v2.4.0, I keep adding functions missing from older Qt versions. It seems, there are many of them which appeared in 5.10 to 5.15. I try to implement them using what I have. However, they are far too numerous and sometimes complex to make all of them. I direct my attention on the simplest and the most used ones. The criterion is quite subjective, I admit.

For now, I'm looking through `QtCore`. I'm closing to the limit of my capabilities, so I won't add many more functions there. I haven't touched `QtGui` and `QtWidgets` yet. Should I add stubs like `NotImplemented` for the missing functions? I sense that it's the best way to warn of an incompatibility, but it's a lot of work.

CAM-Gerlach:
I don't want to put a damper on your hard work here, but we should probably check with @dalthviz and @ccordoba12 before going to the length of fully backporting detailed class-level "polyfills" for complex features like in some of your most recent commits—I just want to make sure that's within scope for QtPy before you do all the work to implement it (and we subsequently have to maintain it), given that like you say doing that for everything is a pretty huge effort. And it might be better to try to land this PR with the low-hanging fruit first and then do a followup to consider more detailed changes, as opposed to trying to cram all of it into one big PR. But maybe not...

@dalthviz , @ccordoba12 , what are your thoughts on this?

StSav012:
The error `AttributeError: type object 'PySide2.QtGui.QColorConstants' has no attribute 'Black'` appeared on `PySide2==5.15.8` on Windows in the latest test is unbelievable! _It's either PySide2 going insane or me._ The attribute is present on Ubuntu, Fedora, and macOS. So report the tests here, and so I see in my tests. As I don't have Windows available to test that very version of PySide2 (nor any other), I've asked the community to double-check; no response so far.

Apart from the error, I've done more or less everything I wanted. As soon as the error is sorted out, the PR is ready, except for one thing. The function dubbed `test_QWizardPage_registerField_changedSignal` from `test_qtwidgets.py` crashes on Ubuntu with `conda=No`. I had to skip the test. I know that I'm not the first to stumble upon errors is that very configuration. However, I can't reproduce the error locally. Can you? If you can, could you please guide me though building the environment that _does_ reproduce the error?

CAM-Gerlach:
> The error `AttributeError: type object 'PySide2.QtGui.QColorConstants' has no attribute 'Black'` appeared on `PySide2==5.15.8` on Windows in the latest test is unbelievable!

Testing on Windows in a brand new clean Conda env with pyside2 5.18.8 and qt-main 5.15.6, I can confirm I see that same issue. `QColorConstants` is present, but does not appear to contain any attributes/enums (including testing lowercase, title case and upper case for both `QColorConstants` and `QColorConstants.Svg`, whereas pyqt 5.15.7 has them and works just fine.

Not sure which approach specifically would be best to patch around that, though.

<details>

<summary>Full output</summary>

```python
Python 3.11.0 | packaged by conda-forge | (main, Oct 25 2022, 06:12:32) [MSC v.1929 64 bit (AMD64)] on win32
>>> import PySide2.QtGui
>>> PySide2.QtGui.QColorConstants
<class 'PySide2.QtGui.QColorConstants'>
>>> PySide2.QtGui.QColorConstants.Black
AttributeError: type object 'PySide2.QtGui.QColorConstants' has no attribute 'Black'
>>> PySide2.QtGui.QColorConstants.White
AttributeError: type object 'PySide2.QtGui.QColorConstants' has no attribute 'White'
>>> dir(PySide2.QtGui.QColorConstants)
['Svg', '__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__getstate__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__']
>>> PySide2.QtGui.QColorConstants.black
AttributeError: type object 'PySide2.QtGui.QColorConstants' has no attribute 'black'
>>> PySide2.QtGui.QColorConstants.Svg
<class 'PySide2.QtGui.QColorConstants.Svg'>
>>> PySide2.QtGui.QColorConstants.Svg.black
AttributeError: type object 'PySide2.QtGui.QColorConstants.Svg' has no attribute 'black'
>>> dir(PySide2.QtGui.QColorConstants.Svg)
['__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__getstate__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__']
>>> PySide2.QtGui.QColorConstants.Svg.Black
AttributeError: type object 'PySide2.QtGui.QColorConstants.Svg' has no attribute 'Black'
>>> PySide2.QtGui.QColorConstants.Svg.BLACK
AttributeError: type object 'PySide2.QtGui.QColorConstants.Svg' has no attribute 'BLACK'
>>> import PyQt5.QtGui
>>> PyQt5.QtGui.QColorConstants
<class 'PyQt5.QtGui.QColorConstants'>
>>> PyQt5.QtGui.QColorConstants.Black
<PyQt5.QtGui.QColor object at 0x00000000037BEA40>
>>> PyQt5.QtGui.QColorConstants.Svg.black
<PyQt5.QtGui.QColor object at 0x00000000037BEAB0>
>>> dir(PyQt5.QtGui.QColorConstants)
['Black', 'Blue', 'Color0', 'Color1', 'Cyan', 'DarkBlue', 'DarkCyan', 'DarkGray', 'DarkGreen', 'DarkMagenta', 'DarkRed', 'DarkYellow', 'Gray', 'Green', 'LightGray', 'Magenta', 'Red', 'Svg', 'Transparent', 'White', 'Yellow', '__class__', '__delattr__', '__dict__', '__dir__', '__doc__', '__eq__', '__format__', '__ge__', '__getattribute__', '__getstate__', '__gt__', '__hash__', '__init__', '__init_subclass__', '__le__', '__lt__', '__module__', '__ne__', '__new__', '__reduce__', '__reduce_ex__', '__repr__', '__setattr__', '__sizeof__', '__str__', '__subclasshook__', '__weakref__']
```

</details>

StSav012:
Thanks, @CAM-Gerlach. The issue looks like a bug, being present only on Windows. I'll make a workaround soon. Is it worth reporting it as a bug to https://bugreports.qt.io/ ? They've been reluctant to develop PySide2 much further.

CAM-Gerlach:
Yeah, I'm not sure you'll make too much headway there due to their recently much more restrictive FOSS-hostile LTS policies. However, I suppose its worth a shot. It could potentially be patched on the Conda side too as well as on ours, of course.

FWIW, I can also confirm I can repro this on non-Conda PySide2 on Windows as well.

dalthviz:
> @dalthviz , @ccordoba12 , what are your thoughts on this?

I think this goes with the idea of providing a uniform layer for Qt bindings indeed. However, seems like a lot to add and review in a single PR, and I'm not sure if we should expand the CI for more specific bindings since the changes not only cover functionality that defears between Qt6 and Qt5 but also between minor Qt5 versions and PyQt vs PySide, right?

I think it could be nice to have the OP description updated with the full list of changes that this PR covers so we have a more clear picture of all the changes when reviewing it.

Also, thank you so much @StSav012 for all the work you have been putting into this!

CAM-Gerlach:
> I think it could be nice to have the OP description updated with the full list of changes that this PR covers so we have a more clear picture of all the changes when reviewing it.

It would also give us a starting point to document this in the future, too.

CAM-Gerlach:
(Sidenote: @dalthviz , if/when this is merged, it should be presumably done via the Squash strategy, or with rebase cleanup first, to avoid nearly a hundred small commits and many reverting back and forth. The squash merge long description will contain a list of the individual changes, and readers can always go refer to the PR for the full commit history of this branch.)

PR #407, opened 2023-02-11
PR: Remove patch that may be limited to PySide-5.11a1 only

StSav012:
The removal looks safe, provided the oldest supported `PySide2` is 5.12.

See https://github.com/spyder-ide/qtpy/issues/405

PR #408, opened 2023-02-13
PR: Add mappings for QMouseEvent methods

StSav012:
Added
```python
    QMouseEvent.pos = lambda self: self.position().toPoint()
    QMouseEvent.x = lambda self: self.position().toPoint().x()
    QMouseEvent.y = lambda self: self.position().toPoint().y()
    QMouseEvent.globalPos = lambda self: self.globalPosition().toPoint()
    QMouseEvent.globalX = lambda self: self.globalPosition().toPoint().x()
    QMouseEvent.globalY = lambda self: self.globalPosition().toPoint().y()
```
for PyQt6 and PySide6.

And I had to add 
```python
    QMouseEvent.position = lambda self: QPointF(float(self.x()), float(self.y()))
    QMouseEvent.globalPosition = lambda self: QPointF(float(self.globalX()), float(self.globalY()))
``` 
to PyQt5 and PySide2 to make the event test universal and (consequently, but this was not the required enhancement) bidirectional. This way, I check that the coordinates returned by the back-ported functions are equal to what the Qt**6** functions return (except for the type, but I use integer coordinates for that to work).

### DIFF
--- a/qtpy/QtCore.py
+++ b/qtpy/QtCore.py
@@ -78,11 +78,6 @@ elif PYQT6:
 elif PYSIDE2:
     from PySide2.QtCore import *
 
-    try:  # may be limited to PySide-5.11a1 only
-        from PySide2.QtGui import QStringListModel
-    except Exception:
-        pass
-
     import PySide2.QtCore
     __version__ = PySide2.QtCore.__version__
 
--- a/qtpy/QtGui.py
+++ b/qtpy/QtGui.py
@@ -73,7 +73,7 @@ if PYSIDE2 or PYSIDE6:
         return movePosition(self, operation, mode, n)
     QTextCursor.movePosition = movePositionPatched
 
-# close https://github.com/spyder-ide/qtpy/issues/394
+# Fix https://github.com/spyder-ide/qtpy/issues/394
 if PYQT6 or PYSIDE6:
     QMouseEvent.pos = lambda self: self.position().toPoint()
     QMouseEvent.x = lambda self: self.position().toPoint().x()
--- a/qtpy/QtGui.py
+++ b/qtpy/QtGui.py
@@ -72,3 +72,12 @@ if PYSIDE2 or PYSIDE6:
     ) -> bool:
         return movePosition(self, operation, mode, n)
     QTextCursor.movePosition = movePositionPatched
+
+# close https://github.com/spyder-ide/qtpy/issues/394
+if PYQT6 or PYSIDE6:
+    QMouseEvent.pos = lambda self: self.position().toPoint()
+    QMouseEvent.x = lambda self: self.position().toPoint().x()
+    QMouseEvent.y = lambda self: self.position().toPoint().y()
+    QMouseEvent.globalPos = lambda self: self.globalPosition().toPoint()
+    QMouseEvent.globalX = lambda self: self.globalPosition().toPoint().x()
+    QMouseEvent.globalY = lambda self: self.globalPosition().toPoint().y()
--- a/qtpy/QtGui.py
+++ b/qtpy/QtGui.py
@@ -74,6 +74,10 @@ if PYSIDE2 or PYSIDE6:
     QTextCursor.movePosition = movePositionPatched
 
 # Fix https://github.com/spyder-ide/qtpy/issues/394
+if PYQT5 or PYSIDE2:
+    from qtpy.QtCore import QPointF as __QPointF
+    QMouseEvent.position = lambda self: __QPointF(float(self.x()), float(self.y()))
+    QMouseEvent.globalPosition = lambda self: __QPointF(float(self.globalX()), float(self.globalY()))
 if PYQT6 or PYSIDE6:
     QMouseEvent.pos = lambda self: self.position().toPoint()
     QMouseEvent.x = lambda self: self.position().toPoint().x()
--- a/qtpy/tests/test_qtgui.py
+++ b/qtpy/tests/test_qtgui.py
@@ -89,6 +89,7 @@ def test_QMouseEvent_pos_functions(qtbot):
             event.accept()
 
     window = Window()
+    window.setMinimumSize(320, 240)  # ensure the window is of sufficient size
     window.show()
 
     QtCore.QTimer.singleShot(100, lambda: qtbot.mouseMove(window, QtCore.QPoint(42, 6 * 9)))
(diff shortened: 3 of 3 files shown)


qtpy/QtGui.py
```diff
@@ -72,3 +72,12 @@ def movePositionPatched(
     ) -> bool:
         return movePosition(self, operation, mode, n)
     QTextCursor.movePosition = movePositionPatched
+
+# close https://github.com/spyder-ide/qtpy/issues/394
```
CAM-Gerlach, line 76:
```suggestion
# Fix https://github.com/spyder-ide/qtpy/issues/394
```

Here we're not actually closing the issue (like e.g. in a PR body), but rather trying to fix it.

CAM-Gerlach (commented):
Seems good so far, thanks, but would it be possible to actually test that the functions can be called correctly and consistently? That seems to be the user's concern on that issue. Thanks!

qtpy/QtGui.py
StSav012: There is no consistent constructor for the event. I mean the constructor changed over the years. Hence, I can compose either separate constructors for Qt versions 5.12, 5.15, and 6.3, or make the event appear naturally. I'm to try the latter approach first.

StSav012:
It looks like the test crashes the process on Ubuntu with `conda=No`. Any thoughts on that? I had to skip the test for the config for now.

And it's can't actually get the coordinates I move the mouse to from the event methods. I can't predict the window border dimensions. Maybe I missed something. Anyway, the mapping I use should prevent any default values (like `QPoint(-1, -1)`) from passing the test.

qtpy/tests/test_qtgui.py
```diff
@@ -61,6 +61,39 @@ def test_enum_access():
...
+
+@pytest.mark.skipif(
+    sys.platform.startswith('linux') and not_using_conda(),
+    reason="Fatal Python error: Aborted on Linux CI when not using conda")
+@pytest.mark.skipif(
+    sys.platform == 'darwin' and sys.version_info[:2] == (3, 7),
+    reason="Stalls on macOS CI with Python 3.7")
+def test_QMouseEvent_pos_functions(qtbot):
+    """Test `QMouseEvent.pos` and related functions obsolete in Qt6,
+       test `QMouseEvent.position` and related functions missing from Qt5"""
```
CAM-Gerlach, line 73: Minor nits, but per the docstring standard and with some minor fixes, this should be

```suggestion
    """
    Test `QMouseEvent.pos` and related functions removed in Qt 6,
    and `QMouseEvent.position`, etc., missing from Qt 5.
    """
```

qtpy/tests/test_qtgui.py
```diff
@@ -61,6 +61,39 @@ def test_enum_access():
...
+            assert event.position().y() == event.pos().y()
+            assert event.globalPosition().x() == event.globalPos().x()
+            assert event.globalPosition().y() == event.globalPos().y()
+
+            event.accept()
+
+    window = Window()
+    window.show()
+
+    QtCore.QTimer.singleShot(100, lambda: qtbot.mouseMove(window, QtCore.QPoint(42, 6 * 9)))
```
CAM-Gerlach, line 92: Nice

CAM-Gerlach (approved):
Looks generally good to me, thanks, aside from another minor nit; for more substantive expert feedback I would defer to @dalthviz

qtpy/tests/test_qtgui.py
StSav012: Now even nicer, except for that nasty `ERROR module 'importlib_metadata' has no attribute 'distributions'`, which started appearing randomly this week. It's not a `QtPy` error.

It turned out that the functions, which control the mouse, can start as soon as the window appears. And I don't ever need to close it.

CAM-Gerlach:
(Sidenote: @dalthviz , if/when this is merged, it should be presumably done via the Squash strategy (or with rebase cleanup first) to avoid all the repeated fixup commits for a relatively small change. I enabled a setting that will still preserve a list of the commit messages in the commit description, and the full commit history of this branch can still be referred to on the PR in perpetuity, while the history on the main branch is kept clean.)

qtpy/tests/test_qtgui.py
CAM-Gerlach: > Now even nicer, except for that nasty `ERROR module 'importlib_metadata' has no attribute 'distributions'`, which started appearing randomly this week. It's not a `QtPy` error.

Presumably something in your environment depends on `importlib_metadata`, but the installed version is too old (and is presumably not constrained properly by the version of that dep you have installed, assuming your env is consistent). If you're using a conda-based distribution, could also be the result of pip-conda mixing or similar issues.

StSav012: It's not in my environment. It's here, on GitHub: https://github.com/spyder-ide/qtpy/actions/runs/4210802198/jobs/7308717859

Later runs have not reproduced the error, although I messed up with the branch, and I'm sorry about it.

CAM-Gerlach: Ah, I see, thanks—must have been a transient transitive dep issue.

qtpy/QtCore.py
```diff
@@ -78,11 +78,6 @@
 elif PYSIDE2:
     from PySide2.QtCore import *
 
-    try:  # may be limited to PySide-5.11a1 only
-        from PySide2.QtGui import QStringListModel
-    except Exception:
-        pass
-
```
dalthviz, line 85: Was this change not done in #407 ? Maybe something happened when updating/merging the local branch with upstream or the remote branch? @CAM-Gerlach you know if this should prevent us from merging this as it is or maybe with the squash merge is not important?

dalthviz (approved):
Thanks @StSav012 for your work here! Besides that seems like there is a change from a PR #407 also here, this LGTM :+1: 

Do you know what could have happened there @CAM-Gerlach? Also, feel free to squash merge this when you think is ready (seems like there a unresolved conversation)

qtpy/QtCore.py
CAM-Gerlach: Yeah, I noticed that; not 100% sure of the exact chain of events without doing a deep-dive investigation, but something went wrong when the contributor merged with the remote branch on their fork, which is messing up the merge-base used for the three-way diff on the PR (as that change was indeed applied to ``master``). I'm almost certain it will vanish when squash-merged and not cause any issues, since all the commits are squashed and that new commit is rebased against the latest ``master``, which is why I didn't say anything before, and if there was a problem GitHub would not allow the merge.

CAM-Gerlach: And yup, looked like it indeed worked as expected :tada:

PR #417, opened 2023-03-24
Fix #394 for all children of `QSinglePointEvent`

### LINKED ISSUE #394: Differences in QEvent subclass APIs in PyQt6 cause attribute and/or type errors
For example, my application uses qtconsole, which in turns uses QtPy. At one point, qtconsole does:
```
        elif etype == QtCore.QEvent.MouseMove:
            anchor = self._control.anchorAt(event.pos())
            QtWidgets.QToolTip.showText(event.globalPos(), anchor)
```
However, when running PyQt6, the method `QMouseEvent.globalPos()` has been removed (as it has also been removed in Qt6 itself). Simply changing to call `QMouseEvent.globalPosition()` doesn't work, as the return type of that is different, `QPoint` (old) v.s. `QPointF` (new).

This is particularly annoying as most event classes used to have the same API - `pos` and `globalPos` returning `QPoint`, sometimes `posF` etc. for a floating version. Now in Python you have to do a load of type checks when you have a method that handles multiple event types. In C++ this would be spotted at compile time, which is probably why the Qt devs thought it was a reasonable change to make.

I would propose mapping this method similar to how e.g. `QFontMetrics.width` is, by monkey patching it directly onto the class object after import. i.e.:

`QMouseEvent.globalPos = lambda self: self.globalPosition().toPoint()`

There may be other missing methods, for example `QMouseEvent.pos()` doesn't actually exist in Qt6 but is defined in PyQt6 anyway - I don't know whether it exists in PySide6.

StSav012:
Analyzing https://github.com/spyder-ide/qtpy/pull/408, I've noticed that the functions I made for Qt6 are not actually of `QMouseEvent`, but belong to its base class `QSinglePointEvent`. So, that's the quick fix for that. I'm sorry I haven't done this several days earlier.

Unfortunately, I couldn't think of good tests for `QNativeGestureEvent`, `QEnterEvent`, `QTabletEvent`, `QHoverEvent`, and `QWheelEvent`, which all are the child classes of `QSinglePointEvent` in Qt6. I'd really appreciate some help here.


StSav012:
Added dummy checks (`assert something is not None`).

As a side note, would [`qtbot.waitExposed`](https://pytest-qt.readthedocs.io/en/latest/reference.html?highlight=waitExposed#pytestqt.qtbot.QtBot.waitExposed) help the stalls of the tests on macOS CI with Python 3.7? The docs read
> This is mainly useful for asynchronous systems like X11, where a window will be mapped to screen some time after being asked to show itself on the screen.

I'd like to try, but not in a PR here.

qtpy/tests/test_qtgui.py
```diff
@@ -101,8 +101,47 @@ def mouseDoubleClickEvent(self, event: QtGui.QMouseEvent) -> None:
...
+    assert QtGui.QHoverEvent.y is not None
+    assert QtGui.QHoverEvent.position is not None
+    assert QtGui.QTabletEvent.pos is not None
+    assert QtGui.QTabletEvent.x is not None
+    assert QtGui.QTabletEvent.y is not None
+    assert QtGui.QTabletEvent.globalPos is not None
+    assert QtGui.QTabletEvent.globalX is not None
+    assert QtGui.QTabletEvent.globalY is not None
+    assert QtGui.QTabletEvent.position is not None
+    assert QtGui.QTabletEvent.globalPosition is not None
```
CAM-Gerlach, line 144: This is getting a bit code smelly. Could you maybe replace this with a two-layer nested `for` loop using `getattr`?

```suggestion
    for name in ['QNativeGestureEvent', 'QEnterEvent', 'QTabletEvent', 'QHoverEvent']:
        for attr in ['pos', 'x', 'y', 'globalPos', 'globalX', 'globalY', 'position', 'globalPosition']:
            assert getattr(QtGui, name) is not None
```

Also, `QTabletEvent` is repeated twice. Did you mean to include something else, like `QSinglePointEvent`?

CAM-Gerlach (commented):
One comment on how to simplify the tests

qtpy/tests/test_qtgui.py
StSav012: Oh, the lines are just an extra copy. Initially, I thought of adding `QWheelEvent` to the list. As for your suggestion, it's not equivalent to my code, although close. I'll fix the code in and hour.

StSav012:
> One comment on how to simplify the tests

That actually is what the review process for. Thanks, @CAM-Gerlach!

One of the tests failed, but it's not me this time. I wish there was a way to stabilize the test environment and reuse it to speed the tests up.

CAM-Gerlach:
I restarted the failed job. If it keeps happening, I can look into it further or explore other options like locked dependencies or additional caching.

qtpy/QtGui.py
```diff
@@ -87,12 +87,33 @@ def movePositionPatched(
...
 if PYQT5 or PYSIDE2:
     from qtpy.QtCore import QPointF as __QPointF
-    QMouseEvent.position = lambda self: __QPointF(float(self.x()), float(self.y()))
+    QNativeGestureEvent.x = lambda self: self.localPos().toPoint().x()
+    QNativeGestureEvent.y = lambda self: self.localPos().toPoint().y()
+    QNativeGestureEvent.position = lambda self: self.localPos()
+    QNativeGestureEvent.globalX = lambda self: self.globalPos().x()
+    QNativeGestureEvent.globalY = lambda self: self.globalPos().y()
+    QNativeGestureEvent.globalPosition = lambda self: __QPointF(float(self.globalPos().x()),
+                                                                float(self.globalPos().y()))
```
ccordoba12, line 96:
```suggestion
    QNativeGestureEvent.globalPosition = lambda self: __QPointF(
        float(self.globalPos().x()), float(self.globalPos().y()))
```

qtpy/QtGui.py
```diff
@@ -87,12 +87,33 @@ def movePositionPatched(
...
+    QEnterEvent.position = lambda self: self.localPos()
+    QEnterEvent.globalPosition = lambda self: __QPointF(float(self.globalX()), float(self.globalY()))
```
ccordoba12, line 98:
```suggestion
    QEnterEvent.globalPosition = lambda self: __QPointF(
        float(self.globalX()), float(self.globalY()))
```

qtpy/QtGui.py
```diff
@@ -87,12 +87,33 @@ def movePositionPatched(
...
+    QTabletEvent.position = lambda self: self.posF()
+    QTabletEvent.globalPosition = lambda self: self.globalPosF()
+    QHoverEvent.x = lambda self: self.pos().x()
+    QHoverEvent.y = lambda self: self.pos().y()
+    QHoverEvent.position = lambda self: self.posF()
+    # no `QHoverEvent.globalPosition`, `QHoverEvent.globalX`, nor `QHoverEvent.globalY` in the Qt5 docs
```
ccordoba12, line 104:
```suggestion
    # No `QHoverEvent.globalPosition`, `QHoverEvent.globalX`,
    # nor `QHoverEvent.globalY` in the Qt5 docs.
```

qtpy/QtGui.py
```diff
@@ -87,12 +87,33 @@ def movePositionPatched(
...
+    QMouseEvent.position = lambda self: self.localPos()
     QMouseEvent.globalPosition = lambda self: __QPointF(float(self.globalX()), float(self.globalY()))
```
ccordoba12, line 106:
```suggestion
    QMouseEvent.globalPosition = lambda self: __QPointF(
        float(self.globalX()), float(self.globalY()))
```

qtpy/tests/test_qtgui.py
```diff
@@ -101,8 +101,16 @@ def mouseDoubleClickEvent(self, event: QtGui.QMouseEvent) -> None:
...
 
-    qtbot.mouseMove(window, QtCore.QPoint(42, 6 * 9))
-    qtbot.mouseDClick(window, QtCore.Qt.LeftButton)
+    with qtbot.waitExposed(window):
+        qtbot.mouseMove(window, QtCore.QPoint(42, 6 * 9))
+        qtbot.mouseDClick(window, QtCore.Qt.LeftButton)
+
+    # the rest of the functions are not actually tested
+    for _class in ('QNativeGestureEvent', 'QEnterEvent', 'QTabletEvent'):
+        for _function in ('pos', 'x', 'y', 'globalPos', 'globalX', 'globalY', 'position', 'globalPosition'):
```
ccordoba12, line 110:
```suggestion
        for _function in ('pos', 'x', 'y', 'globalPos', 'globalX', 'globalY',
                          'position', 'globalPosition'):
```

qtpy/QtGui.py
```diff
@@ -87,12 +87,33 @@ def movePositionPatched(
...
 if PYQT6 or PYSIDE6:
-    QMouseEvent.pos = lambda self: self.position().toPoint()
-    QMouseEvent.x = lambda self: self.position().toPoint().x()
-    QMouseEvent.y = lambda self: self.position().toPoint().y()
-    QMouseEvent.globalPos = lambda self: self.globalPosition().toPoint()
-    QMouseEvent.globalX = lambda self: self.globalPosition().toPoint().x()
-    QMouseEvent.globalY = lambda self: self.globalPosition().toPoint().y()
+    for _class in (QNativeGestureEvent, QEnterEvent, QTabletEvent, QHoverEvent, QMouseEvent):
```
ccordoba12, line 108:
```suggestion
    for _class in (QNativeGestureEvent, QEnterEvent, QTabletEvent, QHoverEvent,
                   QMouseEvent):
```

ccordoba12 (commented):
Thanks @StSav012! I left some small suggestions fro you to avoid very long lines in the code, otherwise looks good to me.

CAM-Gerlach (approved):
LGTM, at least as far as my expertise is concerned :+1:

StSav012:
Thanks @ccordoba12 ! What line length do you prefer? I stick to 120 char limit. I'll gladly follow your code style to save you the extra work on cosmetics.

CAM-Gerlach:
>Thanks, @ccordoba12! What line length do you prefer? I stick to 120 char limit. I'll gladly follow your code style to save you the extra work on cosmetics.

At least in most Spyder projects, nominally 79 characters per PEP 8, but the current code is somewhat inconsistent until we blacken it (and even then, we might want to go with the default 88 characters to avoid too much churn and to avoid too many line breaks on the large number of complex lines QtPy has). I've tried to avoid too many style nitpicks myself except when in the course of another suggestion.

> @CAM-Gerlach, please apply such formatting changes without consulting me to save time (especially when a commit delays a release). Living in Europe, I read the comments many hours after they appear.

Sure, I can do that for you in the future if you like; we just generally like to be very respectful of the author's wishes and not do so without their explicit permission (which you've now given) to avoid any friction and misunderstandings (as has happened in the past). Thanks!

PR #421, opened 2023-04-05
PR: Symmetrize `QDateTime.toPython` and `toPyDateTime`, etc.

StSav012:
As far as I've noticed, the favorable Qt5 binding was PyQt5, and for Qt6 it's PySide6. They are different in many ways beside the Qt backend. One of the differences is the way to get Python date/time from the Qt one: it's `toPyDateTime`/`toPyDate`/`toPyTime` in PyQt5/6 and one `toPython` to rule them all in PySide2/6.

To ease the transfer from PyQt5 to PySide6, I propose the expansion of https://github.com/spyder-ide/qtpy/pull/169. And some tests for the results, as @CAM-Gerlach has suggested in https://github.com/spyder-ide/qtpy/pull/403#discussion_r1095370537.

### DIFF
--- a/qtpy/tests/test_qtcore.py
+++ b/qtpy/tests/test_qtcore.py
@@ -1,7 +1,7 @@
 """Test QtCore."""
 
-from datetime import date, datetime, time
 import sys
+from datetime import datetime
 
 import pytest
 
@@ -16,34 +16,41 @@ from qtpy import (
 )
 from qtpy.tests.utils import not_using_conda
 
+NOW = datetime.now()
+# Make integer milliseconds; `floor` here, don't `round`!
+NOW = NOW.replace(microsecond=(NOW.microsecond // 1000 * 1000))
+
 
 def test_qtmsghandler():
     """Test qtpy.QtMsgHandler"""
     assert QtCore.qInstallMessageHandler is not None
 
 
-def test_qdatetime_toPython():
-    """Test QDateTime.toPython"""
-    q_date = QtCore.QDateTime.currentDateTime()
-    assert QtCore.QDateTime.toPython is not None
+def test_QDateTime_toPython_and_toPyDateTime():
+    """Test `QDateTime.toPython` and `QDateTime.toPyDateTime`"""
+    q_datetime = QtCore.QDateTime(NOW)
+    py_date = q_datetime.toPython()
+    assert py_date == NOW
+    py_date = q_datetime.toPyDateTime()
+    assert py_date == NOW
+
+
+def test_QDate_toPython_and_toPyDate():
+    """Test `QDate.toPython` and `QDate.toPyDate`"""
+    q_date = QtCore.QDateTime(NOW).date()
     py_date = q_date.toPython()
-    assert isinstance(py_date, datetime)
+    assert py_date == NOW.date()
+    py_date = q_date.toPyDate()
+    assert py_date == NOW.date()
 
 
-def test_qdate_toPython():
-    """Test QDate.toPython"""
-    q_date = QtCore.QDate.currentDate()
-    assert QtCore.QDate.toPython is not None
-    py_date = q_date.toPython()
-    assert isinstance(py_date, date)
-
-
-def test_qtime_toPython():
-    """Test QTime.toPython"""
-    q_time = QtCore.QTime.currentTime()
-    assert QtCore.QTime.toPython is not None
+def test_QTime_toPython_and_toPyTime():
+    """Test `QTime.toPython` and `QTime.toPyTime`"""
+    q_time = QtCore.QDateTime(NOW).time()
     py_time = q_time.toPython()
-    assert isinstance(py_time, time)
+    assert py_time == NOW.time()
+    py_time = q_time.toPyTime()
+    assert py_time == NOW.time()
 
 
 @pytest.mark.skipif(
--- a/qtpy/tests/test_qtcore.py
+++ b/qtpy/tests/test_qtcore.py
@@ -1,7 +1,7 @@
 """Test QtCore."""
 
 import sys
-from datetime import datetime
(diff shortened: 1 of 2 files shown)


StSav012:
What's the failed check? How has it come to that conclusion? I guess it's with `conda=Yes`.

Should I split the tests into individual functions? I certainly can, but that would decrease the readability of the code, making numerous functions, of which the `def` lines, the doc strings, and the spaces in between would make the most of the code. Anyway, it's up to you to decide.

qtpy/tests/test_qtcore.py
```diff
@@ -16,34 +16,41 @@
 )
 from qtpy.tests.utils import not_using_conda
 
+NOW = datetime.now()
+# Make integer milliseconds; `floor` here, don't `round`!
+NOW = NOW.replace(microsecond=(NOW.microsecond // 1000 * 1000))
```
CAM-Gerlach, line 21: To avoid double-assignment of a module-level constant, why not

```suggestion
# Make integer milliseconds; `floor` here, don't `round`!
NOW = datetime.now().replace(microsecond=(NOW.microsecond // 1000 * 1000))
```

qtpy/tests/test_qtcore.py
```diff
@@ -16,34 +16,41 @@
...
-    assert QtCore.QDateTime.toPython is not None
-    py_date = q_date.toPython()
-    assert isinstance(py_date, datetime)
+def test_QDateTime_toPython_and_toPyDateTime():
+    """Test `QDateTime.toPython` and `QDateTime.toPyDateTime`"""
+    q_datetime = QtCore.QDateTime(NOW)
+    py_date = q_datetime.toPython()
+    assert py_date == NOW
+    py_date = q_datetime.toPyDateTime()
+    assert py_date == NOW
```
CAM-Gerlach, line 35: I suggest paratermizing each of the tests for the two methods instead of conflating the two and duplicating logic, with the name of the method to call, i.e. something like

```suggestion
@pytest.mark.parametrize("method", ["toPython", "toPyDateTime"])
def test_QDateTime_toPython_and_toPyDateTime(method):
    """Test `QDateTime.toPython` and `QDateTime.toPyDateTime`"""
    q_datetime = QtCore.QDateTime(NOW)
    py_date = getattr(q_datetime, method)()
    assert py_date == NOW
```

And the same for the other tests.

CAM-Gerlach (commented):
In addition to parameterizing the tests, I also suggest doing an `isinstance` check that `py_date` has the expected Python type in each case. Thanks!

CAM-Gerlach:
> What's the failed check? How has it come to that conclusion? I guess it's with `conda=Yes`.

Seems like the coveralls hook status is just misbehaving because the actual run its linked to confirms the status is still the same. Hopefully the next push should fix it.

> Should I split the tests into individual functions? I certainly can, but that would decrease the readability of the code, making numerous functions, of which the `def` lines, the doc strings, and the spaces in between would make the most of the code. Anyway, it's up to you to decide.

As my review suggests, you can instead achieve the same using Pytest's paramertization, which is actually shorter than the existing test code.

qtpy/tests/test_qtcore.py
StSav012: There is the second `NOW` there.

CAM-Gerlach: Oh, d'oh! In that case, you _could_ do...


```suggestion
_now = datetime.now()
# Make integer milliseconds; `floor` here, don't `round`!
NOW = _now.replace(microsecond=(_now.microsecond // 1000 * 1000))
```

qtpy/tests/test_qtcore.py
StSav012: Done: https://github.com/spyder-ide/qtpy/pull/421/commits/553338d443794c662bbe4a8ae8e287cf463535fe

qtpy/tests/test_qtcore.py
StSav012: Done: https://github.com/spyder-ide/qtpy/pull/421/commits/6ee0931475bc2c33c93d9365ed1d51b7a65f11a1

CAM-Gerlach (approved):
LGTM from my side, thanks @StSav012 . @dalthviz ?

StSav012:
> In addition to parameterizing the tests, I also suggest doing an `isinstance` check that `py_date` has the expected Python type in each case.

I'm sorry I forgot to answer this. The point is that the type check is conducted internally in the `__eq__` function. These are the fragments of `datetime.py` (identical for Python 3.7 and 3.11):

- in `datetime.date`:
```python
    def __eq__(self, other):
        if isinstance(other, date):
            return self._cmp(other) == 0
        return NotImplemented
```

- in `datetime.time`:
```python
    def __eq__(self, other):
        if isinstance(other, time):
            return self._cmp(other, allow_mixed=True) == 0
        else:
            return NotImplemented
```

- in `datetime.datetime`:
```python
    def __eq__(self, other):
        if isinstance(other, datetime):
            return self._cmp(other, allow_mixed=True) == 0
        elif not isinstance(other, date):
            return NotImplemented
        else:
            return False
```

CAM-Gerlach:
Ah, I see—technically, `py_date`'s `__eq__` (which gets called first being on the LHS, or if on the RHS would get called if the `NotImplemented` path is taken in the `__eq__` above) could theoretically compare equal with a `datetime`, etc. without being one. But that's probably mostly hypothetical pedantry on my part.

StSav012:
> being on the LHS

Oh, indeed. Thanks for the catch. I've added the explicit type checks. This should make the code more clear.

CAM-Gerlach:
Thanks!

dalthviz (approved):
Thanks @StSav012 !

PR #422, opened 2023-04-05
PR: Use static calls of `exec_` elsewhere where needed, and test them

StSav012:
The help for `QMenu.exec_` reads something like the following (the order might be different, and fully qualified Qt class names might appear):
```plaintext
exec_(...)
    exec_(actions: typing.Sequence[QAction], pos: QPoint, at: typing.Optional[QAction] = None, parent: typing.Optional[QWidget] = None) -> QAction
    exec_(self) -> QAction
    exec_(self, pos: QPoint, at: typing.Optional[QAction] = None) -> QAction
```
The first option here is a static call to `QMenu.exec_`. In the QtPy tests, such a call is never tested. So, here it is. As one might expect, it fails. Use the `_possibly_static_exec` function to fix the errors.

### DIFF
--- a/qtpy/QtWidgets.py
+++ b/qtpy/QtWidgets.py
@@ -9,7 +9,7 @@
 """Provides widget classes and functions."""
 
 from . import PYQT5, PYQT6, PYSIDE2, PYSIDE6, QtModuleNotInstalledError
-from .utils import _possibly_static_exec, getattr_missing_optional_dep
+from .utils import possibly_static_exec, getattr_missing_optional_dep
 
 
 _missing_optional_names = {}
@@ -47,9 +47,9 @@ elif PYQT6:
     QPlainTextEdit.setTabStopWidth = lambda self, *args, **kwargs: self.setTabStopDistance(*args, **kwargs)
     QPlainTextEdit.tabStopWidth = lambda self, *args, **kwargs: self.tabStopDistance(*args, **kwargs)
     QPlainTextEdit.print_ = lambda self, *args, **kwargs: self.print(*args, **kwargs)
-    QApplication.exec_ = lambda *args, **kwargs: _possibly_static_exec(QApplication, *args, **kwargs)
+    QApplication.exec_ = lambda *args, **kwargs: possibly_static_exec(QApplication, *args, **kwargs)
     QDialog.exec_ = lambda self, *args, **kwargs: self.exec(*args, **kwargs)
-    QMenu.exec_ = lambda *args, **kwargs: _possibly_static_exec(QMenu, *args, **kwargs)
+    QMenu.exec_ = lambda *args, **kwargs: possibly_static_exec(QMenu, *args, **kwargs)
     QLineEdit.getTextMargins = lambda self: (self.textMargins().left(), self.textMargins().top(), self.textMargins().right(), self.textMargins().bottom())
 
     # Allow unscoped access for enums inside the QtWidgets module
@@ -82,6 +82,6 @@ elif PYSIDE6:
     QLineEdit.getTextMargins = lambda self: (self.textMargins().left(), self.textMargins().top(), self.textMargins().right(), self.textMargins().bottom())
 
     # Map DeprecationWarning methods
-    QApplication.exec_ = lambda *args, **kwargs: _possibly_static_exec(QApplication, *args, **kwargs)
+    QApplication.exec_ = lambda *args, **kwargs: possibly_static_exec(QApplication, *args, **kwargs)
     QDialog.exec_ = lambda self, *args, **kwargs: self.exec(*args, **kwargs)
-    QMenu.exec_ = lambda *args, **kwargs: _possibly_static_exec(QMenu, *args, **kwargs)
+    QMenu.exec_ = lambda *args, **kwargs: possibly_static_exec(QMenu, *args, **kwargs)
--- a/qtpy/QtWidgets.py
+++ b/qtpy/QtWidgets.py
@@ -9,7 +9,7 @@
 """Provides widget classes and functions."""
 
 from . import PYQT5, PYQT6, PYSIDE2, PYSIDE6, QtModuleNotInstalledError
-from .utils import getattr_missing_optional_dep
+from .utils import _possibly_static_exec, getattr_missing_optional_dep
 
 
 _missing_optional_names = {}
@@ -19,17 +19,6 @@ def __getattr__(name):
     """Custom getattr to chain and wrap errors due to missing optional deps."""
     raise getattr_missing_optional_dep(
         name, module_name=__name__, optional_names=_missing_optional_names)
-
-
-def _possibly_static_exec(cls, *args, **kwargs):
-    """Call `self.exec` when `self` is given or a static method otherwise."""
-    if isinstance(args[0], cls):
-        if len(args) == 1 and not kwargs:
-            # A special case to avoid the function resolving error
-            return args[0].exec()
-        return args[0].exec(*args[1:], **kwargs)
-    else:
-        return cls.exec(*args, **kwargs)
 
 
 if PYQT5:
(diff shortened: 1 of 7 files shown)


StSav012:
The mistake of mine was in the following lines:
```python
        QtCore.QTimer.singleShot(100, lambda: qtbot.keyClick(
            QtWidgets.QApplication.widgetAt(1, 1),
            QtCore.Qt.Key.Key_Escape)
            # namely, ↑↑↑ up here
        )
```
In `PyQt5=5.9`, there was no `QtCore.Qt.Key.Key_Escape` yet. Surprisingly, the attribute error was suppressed, and no timer shot was fired. That caused the stall. When I changed the parameter to `QtCore.Qt.Key_Escape`, it worked.

As I don't use Windows, and the problem appears only on Windows, I couldn't catch the problem locally. The third time's a charm.

StSav012:
Oh, it's `ERROR module 'importlib_metadata' has no attribute 'distributions'` again. I see the error is so persistent that @dalthviz has merged https://github.com/spyder-ide/qtpy/commit/df95964f87a5956c4b97005f5fe6069904901b43 into master despite the failed check.

CAM-Gerlach:
> Oh, it's `ERROR module 'importlib_metadata' has no attribute 'distributions'` again.

Thanks, I opened #423 to document that.

CAM-Gerlach:
(Fixed the `importlib` error in #425 , BTW, while also speeding up the Conda=yes CI jobs by roughly 2x and simplifying the CI script a bit.)

qtpy/QtWidgets.py
```diff
@@ -14,12 +14,24 @@
...
 
+
 def __getattr__(name):
     """Custom getattr to chain and wrap errors due to missing optional deps."""
     raise getattr_missing_optional_dep(
         name, module_name=__name__, optional_names=_missing_optional_names)
 
 
+def _possibly_static_exec(cls, *args, **kwargs):
+    """ Call `self.exec` when `self` is given or a static method otherwise. """
```
CAM-Gerlach, line 25:
```suggestion
    """Call `self.exec` when `self` is given or a static method otherwise."""
```

Nit: Strip excess white space per PEP 257.

CAM-Gerlach (commented):
Sorry, didn't realize I hadn't actually reviewed this.

If it might be used in multiple modules, maybe better to move `_possibly_static_exec` to `utils`? Thoughts?

Beyond that, I didn't spot any obvious blocking issues with this, so otherwise leaving this to @dalthviz 's expertise to review further.

StSav012:
> If it might be used in multiple modules, maybe better to move `_possibly_static_exec` to `utils`?

I've looked through 
- `QApplication` ([Qt5](https://doc.qt.io/qt-5/qapplication.html#static-public-members), [Qt6](https://doc.qt.io/qt-6/qapplication.html#static-public-members)),
- `QCoreApplication` ([Qt5](https://doc.qt.io/qt-5/qcoreapplication.html#static-public-members), [Qt6](https://doc.qt.io/qt-6/qcoreapplication.html#static-public-members)),
- `QDialog` ([Qt5](https://doc.qt.io/qt-5/qdialog.html#static-public-members), [Qt6](https://doc.qt.io/qt-6/qdialog.html#static-public-members)), 
- `QDrag` ([Qt5](https://doc.qt.io/qt-5/qdrag.html#static-public-members), [Qt6](https://doc.qt.io/qt-6/qdrag.html#static-public-members)),
- `QEventLoop` ([Qt5](https://doc.qt.io/qt-5/qeventloop.html#static-public-members), [Qt6](https://doc.qt.io/qt-6/qeventloop.html#static-public-members)), 
- `QGuiApplication` ([Qt5](https://doc.qt.io/qt-5/qguiapplication.html#static-public-members), [Qt6](https://doc.qt.io/qt-6/qguiapplication.html#static-public-members)),
- `QMenu` ([Qt5](https://doc.qt.io/qt-5/qmenu.html#static-public-members), [Qt6](https://doc.qt.io/qt-6/qmenu.html#static-public-members)), 
- `QTextStreamManipulator` ([PySide2](https://doc.qt.io/qtforpython-5/PySide2/QtCore/QTextStreamManipulator.html#PySide2.QtCore.PySide2.QtCore.QTextStreamManipulator.exec_), [PySide6](https://doc.qt.io/qtforpython/PySide6/QtCore/QTextStreamManipulator.html)),
-  and `QThread` ([Qt5](https://doc.qt.io/qt-5/qthread.html#static-public-members), [Qt6](https://doc.qt.io/qt-6/qthread.html#static-public-members)), 

and found that `QCoreApplication`, `QGuiApplication`, `QApplication`, and `QMenu` indeed have such static calls. Thank you for the remark!

As for `QTextStreamManipulator`, I can't compose a test call. I just can't make an instance of `QTextStreamManipulator` last. According to `help(QTextStreamManipulator.exec_)`, it isn't static, but still. That's an odd class.

StSav012:
Surprisingly, the tests for `QtCore.QCoreApplication.exec_` don't result in a segfault on Ubuntu with `conda=No`.

CAM-Gerlach (approved):
Thanks! All looks pretty reasonable to me now, but I'll leave it to @dalthviz 's greater expertise to review further, thanks.

qtpy/utils.py
```diff
@@ -30,3 +30,17 @@ def getattr_missing_optional_dep(name, module_name, optional_names):
     if name in optional_names:
         return _wrap_missing_optional_dep_error(attr_error, **optional_names[name])
     return attr_error
+
+
+def _possibly_static_exec(cls, *args, **kwargs):
```
ccordoba12, line 35: I think this should be renamed to `possibly_static_exec` (just as with `getattr_missing_optional_dep`) because it's called from other modules here, so it's basically a public function.

StSav012: Does it then get exported, like `parse`? It isn't desired behavior. I can delete it at the end, though.

StSav012: Done: https://github.com/spyder-ide/qtpy/pull/422/commits/b59f7c0ba2803660df63ef158751fa0b31fefd3a

CAM-Gerlach: As I commented previously in the `getattr_missing_optional_dep` case, for a library (like QtPy) used by many downstream applications, "private"/"internal" _to the package_ (i.e. QtPy, it is not intended for outside consumers of the library to use) is the much more relevant (and important) meaning here than "private"/"internal" _to the module_ (as would be the case for an application like Spyder):

> It's private to the _package_ as a whole rather than the specific module its defined in. At present, this means it is exposed as a module-level public function in any of the modules its imported in. Furthermore, its not intended to be use by anything except as an internal implementation detail of the common `__getattr__` implementation.

Therefore, like the `getattr` helper, IMO it is entirely appropriate that it be `_` private, as it is only intended for strictly internal use rather than being exposed to users.

Perhaps, though, we could just mark the `utils` module private instead, since was specifically created for such "private"/"internal" utils.

dalthviz (approved):
Thanks @StSav012 ! LGTM :+1: But, just in case, what do you think @ccordoba12 ?

qtpy/utils.py
ccordoba12: > Perhaps, though, we could just mark the utils module private instead (i.e. _utils), since was specifically created for such "private"/"internal" utils.

Yeah, I think that's better. And then rename all functions in it to be public if they are use in other QtPy modules?

StSav012: Should I do it in the PR?

CAM-Gerlach: > And then rename all functions in it to be public if they are used in other QtPy modules?

In this case there's only the two functions, the `possibly_static_exec` added in this PR and the `getattr_missing_optional_deps` just added in another recent PR in which the module was created, and they are generally imported in the same files, so it wouldn't require any function renaming, just fixing a few imports.

> Should I do it in the PR?

I would recommend another PR, to avoid the file move obscuring the additions here.

ccordoba12: > so it wouldn't require any function renaming, just fixing a few imports.

There are only three functions in `utils.py` (the other one being `_wrap_missing_optional_dep_error`), so I think we should rename them all in the same PR.

> I would recommend another PR, to avoid the file move obscuring the additions here.

Agreed.

ccordoba12 (approved):
Thanks @StSav012!

Issue #423, opened 2023-04-08
Fix flaky importlib error on Ubuntu Python 3.7 conda CI job


CAM-Gerlach:
The Test ubuntu-latest Python 3.7 conda=Yes CI job fails approximately every other time with the error `ERROR module 'importlib_metadata' has no attribute 'distributions'`: [Example Run](https://github.com/spyder-ide/qtpy/actions/runs/4636864288/jobs/8205180511?pr=422). We should probably look into that and do something about it.

ccordoba12:
3.7 is about to reach end-of-life, so I'd suggest to drop it from our CIs instead of investing time trying to fix this

CAM-Gerlach:
After looking in to this further I found conda/conda#12390 which explains what's going on; basically, the issue is that the Conda solver solution can sometimes be unstable and not fully accurate when only using `repodata_current.json`; in this case it's happening to the Python 3.7 job, but it could theoretically happen to others as well, or if we bumped the version.

However, the solution is to just use the Mamba solver instead, which not only ensures a more stable solve, but speeds up the (currently slower) `conda=yes` jobs by an average of approximately 2x (especially when combined with some other optimizations, particularly doing a one-pass env creation and installation rather than installing the Qt bindings separately, which also further improves stability and reduces the chance for issues). Plus the various tweaks end up making our CI scripts less duplicative and more maintainable too.

PR #425, opened 2023-04-09
PR: Use Mamba to greatly speed up CI jobs and avoid flaky resolution issues

CAM-Gerlach:
As originally prompted by the fix recommended in conda/conda#12390 for the CI issue in #423 , this will use Mamba rather than Conda on the CIs, which both has more stable, accurate solves avoiding that issue, but also should hopefully run much quicker (since most of the CI time is spent performing Conda operations, particularly solving, that Mamba should dramatically speed up).

Also, try to avoid the somewhat hacky need to manually activate Conda in the CI script, which results in much noisier and hard to read run output.


ccordoba12 (approved):

PR #431, opened 2023-05-17
PR: Rename `utils.py` to `_utils.py` to make it private

StSav012:
Following the https://github.com/spyder-ide/qtpy/pull/422#discussion_r1167280037, I renamed `utils.py`.

As for renaming the `_wrap_missing_optional_dep_error` function, I'm unsure. It's only usage is in the `getattr_missing_optional_dep` function, which is in the same file and immediately follows the former. The underscore conveniently emphasizes that the `_wrap_missing_optional_dep_error` function is not used outside the file. I'd rather stick to @DaelonSuzuka's decision.


dalthviz (approved):
Thanks @StSav012 ! LGTM :+1:

StSav012:
Please forgive me for being impatient, but it's been 45 days since last activity here. The change I propose is trivial. I understand that everyone's busy with the work they're paid for.

I have another PR (about `Q*.exec()`) almost ready, but it will conflict with this one. In a little more than a week, I'm going on vacation for 4 weeks, and I'd like to submit the next PR before that. I sincerely believe that someone might find the next PR useful.

dalthviz:
Sorry @StSav012 ! This fall through the cracks 😅 Merging!

Issue #345, opened 2022-05-02
Use pre-commit to automatically format code with Black and imports with isort


CAM-Gerlach:
As suggested by @DaelonSuzuka in #344 , 

> It would also help readability and auditability to standardize the formatting of these files.

Indeed, for a while now the QtPy source has followed an inconsistent smattering of different PEP 8 and non-PEP 8-conforming styles and practices, which make it harder to easily read and edit, and creates extra work for maintainers and contributors to determine and conform to a consistent code style, as well as extra diff noise and merge conflict potential. This also results in a lot of noisy comments in PR reviews for issues that are primary style-related. This is particularly true since much of QtPy's "code" is import statements for the various names from the different bindings, and given the limited documentation, its particularly important for the code, particularly the imports, to be easily readable for others.

To both improve import and code readability and consistency, and avoid contributors and maintainers having to spend time checking, implementing and enforcing code style, I suggest adding Black to format the code and isort to automatically organize the imports (including conforming to our own custom headings and ordering rules). Adding them via pre-commit, as we do for a number of our other repos, makes it easy for maintainers and contributors to run them automatically whenever they commit, or on-demand, without having to worry about installing, configuring and manually running tooling, and our CIs can use the same as a check. They can also run them individually, or via editor/IDE integration (including in Spyder), if they prefer.

I've already configured this in other projects, so its mostly a matter of dropping in the existing config plus any QtPy-specific config customization. Since we only have one open PR, #344 , aside from a several-year old, stale and heavily conflicting one that would need to be redone anyway, this should be minimally disruptive. I can add the contributor documentation for running pre-commit from our other repos; it basically just boils down to installing pre-commit (if its not already a dev dependency) and running `pre-commit install`, and everything else (including installing, updating and running hooks) requires zero additional action on the part of the user.

dalthviz:
Some suggestions on this topic: https://github.com/spyder-ide/qtpy/pull/433#issuecomment-1611971932

Czaki:
one of latest changes is that it is better to use https://github.com/psf/black-pre-commit-mirror instead of https://github.com/psf/black in pre-commit configuration as first uses pre-compiled wheels and is about 4 times faster.

Czaki:
@dalthviz I see that you have `.pre-commit-config.yaml` with isort and black. What is the real task? Lack of CI setup?  
To big configuration (tom any checks)?

dalthviz:
[...]

PR #433, opened 2023-05-24
PR: Add wrapers to fix argument name in `QFileDialog` `get*`methods

### LINKED ISSUE #432: Add compatibility layer for keywords arguments of `QFileDialog ` `get*` class methods
The `QfileDialog` class contains few static methods for fast opening files: https://doc.qt.io/qt-6/qfiledialog.html#static-public-members

Unfortunately, PyQt and PySide backends differ in the naming argument that provides default directory. 
In PyQt5/6 it is named `directory`, when in PySide2/6 it is named `dir`. PySide follows original Qt convention where PqYt decided to not use the keyword as name of an argument.

Unfortunately, I do not know the method of testing that could find this problem. AS calling this function opened dialogs that hang tests, and with mocking the problem with the signature is not triggered. 

Are you open for PR that will introduce cross-compatibility layer. 

Something like:

```python
if "dir" in kwargs:
    kwargs["directory"] = kwargs.pop("dir")
```

and vice versa? 

inspired by discussion from https://github.com/napari/napari/pull/5834

Czaki:
closes #432

What did you think about this as idea?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,10 +100,6 @@ jobs:
           sudo apt update
           sudo apt install libpulse-dev libegl1-mesa libopengl0 gstreamer1.0-gl
       - uses: tlambert03/setup-qt-libs@v1
-      - name: Install Linux system packages
-        if: contains(matrix.os, 'ubuntu')
-        shell: bash
-        run: sudo apt install libpulse-dev libegl1-mesa libopengl0 gstreamer1.0-gl
       - name: Install Conda
         uses: conda-incubator/setup-miniconda@v2
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -100,6 +100,10 @@ jobs:
           sudo apt update
           sudo apt install libpulse-dev libegl1-mesa libopengl0 gstreamer1.0-gl
       - uses: tlambert03/setup-qt-libs@v1
+      - name: Install Linux system packages
+        if: contains(matrix.os, 'ubuntu')
+        shell: bash
+        run: sudo apt install libpulse-dev libegl1-mesa libopengl0 gstreamer1.0-gl
       - name: Install Conda
         uses: conda-incubator/setup-miniconda@v2
         with:
--- a/qtpy/tests/test_qtwidgets.py
+++ b/qtpy/tests/test_qtwidgets.py
@@ -145,11 +145,11 @@ def test_opengl_imports():
 @pytest.mark.parametrize("instance", [True, False])
 def test_qfiledialog_compat(tmp_path, qtbot, keyword, instance):
     """
-    This function is testing if the decorators that renamed keyword are
-    working.
+    This function is testing if the decorators that renamed the dir/directory
+    keyword are working.
 
-    It mas stop working if the Qt bindings do some overwriting of the methods
-    in constructor. It should not happen, but we already meet that PySide team
+    It may stop working if the Qt bindings do some overwriting of the methods
+    in constructor. It should not happen, but the PySide team
     did similar things in the past (like overwriting enum module in
     PySide6==6.3.2).
 
--- a/qtpy/tests/test_qtwidgets.py
+++ b/qtpy/tests/test_qtwidgets.py
@@ -4,11 +4,11 @@ import sys
 from time import sleep
 
 import pytest
+from pytestqt.exceptions import TimeoutError
 
 from qtpy import PYQT5, PYQT_VERSION, QtCore, QtGui, QtWidgets, PYQT6, PYSIDE6, PYSIDE2
 from qtpy.tests.utils import using_conda, not_using_conda
 
-from pytestqt.exceptions import TimeoutError
 
 
 def test_qtextedit_functions(qtbot, pdf_writer):
@@ -136,29 +136,38 @@ def test_opengl_imports():
     """
     assert QtWidgets.QOpenGLWidget is not None
 
-@pytest.mark.skipif(sys.platform == 'darwin' and sys.version_info[:2] == (3, 7) and (PYQT5 or PYSIDE2), reason="test crashes on macOS CI with Python 3.7 with 'Illegal instruction: 4'")
+
+@pytest.mark.skipif(
+    sys.platform == 'darwin' and sys.version_info[:2] == (3, 7) and
+    (PYQT5 or PYSIDE2),
+    reason="Crashes on macOS with Python 3.7 with 'Illegal instruction: 4'")
 @pytest.mark.parametrize("keyword", ["dir", "directory"])
 @pytest.mark.parametrize("instance", [True, False])
 def test_qfiledialog_compat(tmp_path, qtbot, keyword, instance):
     """
-    This function is testing if the decorators that renamed keyword are working.
-    It mas stop working if the Qt bindings do some overwriting of the methods in constructor. 
-    It should not happen, but we already meet that PySide team did similar things in the past 
-    (like overwriting enum module in PySide6==6.3.2).
+    This function is testing if the decorators that renamed keyword are
+    working.
+
+    It mas stop working if the Qt bindings do some overwriting of the methods
+    in constructor. It should not happen, but we already meet that PySide team
+    did similar things in the past (like overwriting enum module in
+    PySide6==6.3.2).
 
     keyword: str
         The keyword that should be used in the function call.
     instance: bool
-        If True, the function is called on the instance of the QFileDialog, otherwise on the class.
+        If True, the function is called on the instance of the QFileDialog,
+        otherwise on the class.
     """
     
     class CloseThread(QtCore.QThread):
         """
         On some implementations the `getExistingDirectory` functions starts own 
-        event loop that will not trigger QTimer started before the call. Until the
-        dialog is closed the main event loop will be stopped.
+        event loop that will not trigger QTimer started before the call. Until
+        the dialog is closed the main event loop will be stopped.
 
-        Because of this it is required to use the thread to interact with the dialog.
+        Because of this it is required to use the thread to interact with the
+        dialog.
         """
         def run(self, allow_restart=True):
             sleep(0.5)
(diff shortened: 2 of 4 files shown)


dalthviz:
Thanks @Czaki for your work here! Seems like we can go with this approach. Then the only missing thing here would be to try to add a test for this but other than that this LGTM :+1:

Czaki:
After last PySide6 problems I need find way to test it.

Czaki:
How to deal with 
```
UserWarning: Existing QApplication <PySide6.QtCore.QCoreApplication(0x55e7bf91d090) at 0x7f4563b26b70> is not an instance of qapp_cls: <class 'PySide6.QtWidgets.QApplication'>
```

dalthviz:
> UserWarning: Existing QApplication <PySide6.QtCore.QCoreApplication(0x55e7bf91d090) at 0x7f4563b26b70> is not an instance of qapp_cls: <class 'PySide6.QtWidgets.QApplication'>

My guess is that that could come from using the `qapp` fixture maybe? Or where are you seeing that warning?

Also, for the test, maybe using a `QTimer` singleshot instead of a `QThread` to handle the dialog could help? Or maybe the thread code should run inside a `while True` block which breaks when an instance of `QFileDialog` is found and closed?

Czaki:
> Also, for the test, maybe using a `QTimer` singleshot instead of a `QThread`

No. Timers are implemented in Qt as events put in the main queue event (for performance). Especially in documentation, there is: 

> On Windows, the dialog will spin a blocking modal event loop that will not dispatch any QTimers, and if parent is not nullptr then it will position the dialog just below the parent's title bar.

I have also observed problems with timers in such scenarios in the past. It happens that they are not triggered on CI when another event loop is started. 

There is a check if the thread is closed before the end of the new test. 

> My guess is that that could come from using the `qapp` fixture maybe? Or where are you seeing that warning?

Yes but I do not know where it is created.

dalthviz:
>> My guess is that that could come from using the qapp fixture maybe? Or where are you seeing that warning?
>
> Yes but I do not know where it is created.

Seems like there is a fixture that is called `qapp_cls`.  Maybe the warning is related with it? https://pytest-qt.readthedocs.io/en/latest/_modules/pytestqt/plugin.html#qapp_cls

dalthviz:
> It looks like source of all problems was lack of some libraries in system. It looks like it is possible to unskip more tests.
> 
> Are you think it is worth?

Oh that's good to know! I think it would be great if we can unskip tests!

Czaki:
> Oh that's good to know! I think it would be great if we can unskip tests!

done in #434

Czaki:
It is ready. I prefer to fix #434 first and then resolve conflicts.

qtpy/tests/test_qtwidgets.py
```diff
@@ -130,3 +135,74 @@ def test_opengl_imports():
     QtPy makes QOpenGLWidget available in QtWidgets to maintain compatibility.
     """
     assert QtWidgets.QOpenGLWidget is not None
+
+@pytest.mark.skipif(sys.platform == 'darwin' and sys.version_info[:2] == (3, 7) and (PYQT5 or PYSIDE2), reason="test crashes on macOS CI with Python 3.7 with 'Illegal instruction: 4'")
```
ccordoba12, line 139:
```suggestion

@pytest.mark.skipif(
    sys.platform == 'darwin' and sys.version_info[:2] == (3, 7) and
    (PYQT5 or PYSIDE2),
    reason="Crashes on macOS with Python 3.7 with 'Illegal instruction: 4'")
```

This line was super long. We prefer lines to less than 80 columns wide.

qtpy/tests/test_qtwidgets.py
```diff
@@ -1,9 +1,14 @@
 """Test QtWidgets."""
+import contextlib
 import sys
+from time import sleep
 
 import pytest
```
ccordoba12, line 6:
```suggestion
import pytest
from pytestqt.exceptions import TimeoutError
```

`pytestqt` is a third-party import.

qtpy/tests/test_qtwidgets.py
```diff
@@ -1,9 +1,14 @@
...
 
-from qtpy import PYQT5, PYQT_VERSION, QtCore, QtGui, QtWidgets
+from qtpy import PYQT5, PYQT_VERSION, QtCore, QtGui, QtWidgets, PYQT6, PYSIDE6, PYSIDE2
+from qtpy.tests.utils import using_conda, not_using_conda
+
+from pytestqt.exceptions import TimeoutError
```
ccordoba12, line 11:
```suggestion
```

Moved above.

qtpy/tests/test_qtwidgets.py
```diff
@@ -130,3 +135,74 @@ def test_opengl_imports():
...
+    """
+    This function is testing if the decorators that renamed keyword are working.
+    It mas stop working if the Qt bindings do some overwriting of the methods in constructor. 
+    It should not happen, but we already meet that PySide team did similar things in the past 
+    (like overwriting enum module in PySide6==6.3.2).
+
+    keyword: str
+        The keyword that should be used in the function call.
+    instance: bool
+        If True, the function is called on the instance of the QFileDialog, otherwise on the class.
```
ccordoba12, line 152:
```suggestion
    This function is testing if the decorators that renamed keyword are
    working.

    It mas stop working if the Qt bindings do some overwriting of the methods
    in constructor. It should not happen, but we already meet that PySide team
    did similar things in the past (like overwriting enum module in
    PySide6==6.3.2).

    keyword: str
        The keyword that should be used in the function call.
    instance: bool
        If True, the function is called on the instance of the QFileDialog,
        otherwise on the class.
```

Fix formatting to be less than 80 columns wide.

qtpy/tests/test_qtwidgets.py
```diff
@@ -130,3 +135,74 @@ def test_opengl_imports():
...
+    """
+    
+    class CloseThread(QtCore.QThread):
+        """
+        On some implementations the `getExistingDirectory` functions starts own 
+        event loop that will not trigger QTimer started before the call. Until the
+        dialog is closed the main event loop will be stopped.
+
+        Because of this it is required to use the thread to interact with the dialog.
```
ccordoba12, line 161:
```suggestion
        On some implementations the `getExistingDirectory` functions starts own 
        event loop that will not trigger QTimer started before the call. Until
        the dialog is closed the main event loop will be stopped.

        Because of this it is required to use the thread to interact with the
        dialog.
```

80 columns.

qtpy/tests/test_qtwidgets.py
```diff
@@ -130,3 +135,74 @@ def test_opengl_imports():
...
+            sleep(0.5)
+            need_restart = allow_restart
+            app = QtWidgets.QApplication.instance()
+            for dlg in app.topLevelWidgets():
+                if not isinstance(dlg, QtWidgets.QFileDialog) or dlg.isHidden():
+                    continue
+                # when implement this I try to use:
+                # * dlg.close() - On Qt6 it will close the dialog, but it will not restart the main event loop
+                # * dlg.accept() - It ends with information thar `accept` and `reject` of such created dialog can not be called
+                # * accept dialog with enter - It works, but it cannot be called to early after dialog is shown
```
ccordoba12, line 173:
```suggestion
                # "when implemented this I try to use:
                # * dlg.close() - On Qt6 it will close the dialog, but it will
                #   not restart the main event loop.
                # * dlg.accept() - It ends with information thar `accept` and
                #   `reject` of such created dialog can not be called.
                # * accept dialog with enter - It works, but it cannot be
                #   called to early after dialog is shown
```

80 columns

qtpy/tests/test_qtwidgets.py
```diff
@@ -130,3 +135,74 @@ def test_opengl_imports():
...
+                # in such case we call `run` function again. The 0.5s sleep is enough for the second enter to close the dialog. 
+                if not isinstance(dlg, QtWidgets.QFileDialog) or dlg.isHidden():
+                    continue
+                self.run(allow_restart=False)
+                return
+
+            if need_restart:
+                self.run()
+
+    # we need to use the `DontUseNativeDialog` option to be able to interact with it from code.
```
ccordoba12, line 188:
```suggestion
    # We need to use the `DontUseNativeDialog` option to be able to interact
    # with it from code. 
```

qtpy/tests/test_qtwidgets.py
```diff
@@ -130,3 +135,74 @@ def test_opengl_imports():
...
+    }
+
+
+    thr = CloseThread()
+    thr.start()
+    qtbot.waitUntil(thr.isRunning, timeout=1000)
+    dlg = QtWidgets.QFileDialog() if instance else QtWidgets.QFileDialog
+    dlg.getExistingDirectory(**kwargs)
+    qtbot.waitUntil(thr.isFinished, timeout=3000)
+
```
ccordoba12, line 208:
```suggestion
```

Leave eol at the end.

ccordoba12 (commented):
Some minor stylistic suggestions for you @Czaki. I'll let @dalthviz review the code.

Czaki:
Too many projects with black and ruff in pre-commit when there is no need to check formatting in base case as it will be auto fixed.

ccordoba12:
No worries. We're actually behind the times on that front but the thing is we have a lot of code to maintain and little time to do that sort of devops work.

Do you know some simple to set up GH actions to get started with Black and Ruff?

.github/workflows/ci.yml
```diff
@@ -100,6 +100,10 @@ jobs:
           sudo apt update
           sudo apt install libpulse-dev libegl1-mesa libopengl0 gstreamer1.0-gl
       - uses: tlambert03/setup-qt-libs@v1
+      - name: Install Linux system packages
+        if: contains(matrix.os, 'ubuntu')
+        shell: bash
+        run: sudo apt install libpulse-dev libegl1-mesa libopengl0 gstreamer1.0-gl
```
dalthviz, line 106: I think this is duplicating line 101. Maybe the step at lines 96 to 101 need to be after the step that uses the `tlambert03/setup-qt-libs` action for things to work? If the step command duplication is necessary, could we add a comment clarifying why this is needed and maybe rename the step so we don't have two steps in the job with the same name?

qtpy/tests/test_qtwidgets.py
```diff
@@ -130,3 +135,86 @@ def test_opengl_imports():
...
+    reason="Crashes on macOS with Python 3.7 with 'Illegal instruction: 4'")
+@pytest.mark.parametrize("keyword", ["dir", "directory"])
+@pytest.mark.parametrize("instance", [True, False])
+def test_qfiledialog_compat(tmp_path, qtbot, keyword, instance):
+    """
+    This function is testing if the decorators that renamed keyword are
+    working.
+
+    It mas stop working if the Qt bindings do some overwriting of the methods
+    in constructor. It should not happen, but we already meet that PySide team
```
dalthviz, line 152: Little suggestion for this docstring:

```suggestion
    This function is testing if the decorators that renamed the dir/directory
    keyword are working.

    It may stop working if the Qt bindings do some overwriting of the methods
    in constructor. It should not happen, but the PySide team
```

dalthviz (commented):
Thanks @Czaki for all your work here! Left some questions in the workflow file regarding the new step since it seems like it duplicates partially a previous step. Also left a small suggestion for the test docstring. Other than that this LGTM :+1:

On the other hand, regarding the autoformating/pre-commit/black/etc, that's something we need to do indeed! In fact, we have an issue tracking that: https://github.com/spyder-ide/qtpy/issues/345 Also, thanks for the suggestions! I will put a reference to your comment in the issue so we can check them in the future :+1:

.github/workflows/ci.yml
Czaki, line 106:
```suggestion
```

Czaki: No. This is an artifact from rebasing that I did not notice.

dalthviz (approved):
Thanks @Czaki !

PR #434, opened 2023-06-03
PR: Enable more tests on CI

Czaki:
In this PR I have added installation of libs required by qt to setup `QApplication`. For this, I have used [`tlambert03/setup-qt-libs@v1` ](https://github.com/tlambert03/setup-qt-libs). This action is used in few projects testing Qt (like napari) so if new Qt version requires some additional library it will be added without the need to make such a patch in all projects.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,10 @@ on:
     - main
     - '*.x'
 
+concurrency:
+  group: test-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   test:
     name: Test ${{ matrix.os }} Python ${{ matrix.python-version }} conda=${{ matrix.use-conda }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,6 +95,7 @@ jobs:
         run: |
           sudo apt update
           sudo apt install libpulse-dev libegl1-mesa libopengl0 gstreamer1.0-gl
+      - uses: tlambert03/setup-qt-libs@v1
       - name: Install Conda
         uses: conda-incubator/setup-miniconda@v2
         with:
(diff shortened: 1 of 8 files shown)


Czaki:
I have hope that I found all test to uncomment.

dalthviz (commented):
Thanks @Czaki for working on this! Checking seems like maybe there is just one more test which skip on linux could be removed:

https://github.com/spyder-ide/qtpy/blob/1874fbac7ea6c56eb6de58d7d8dceae5229644e0/qtpy/tests/test_compat.py#L12

However, just in case, what do you think @CAM-Gerlach and @ccordoba12 ?

Czaki:
I have removed it and all Linux test pass :)

.github/workflows/ci.yml
```diff
@@ -95,6 +99,7 @@ jobs:
         run: |
           sudo apt update
           sudo apt install libpulse-dev libegl1-mesa libopengl0 gstreamer1.0-gl
```
ccordoba12, line 101: Is still necessary to run these commands with the new `setup-qt-libs` action? Probably not, but just checking.

ccordoba12 (commented):
Left a small question, otherwise looks good to me.

.github/workflows/ci.yml
Czaki: Yes. I need to identify which one is missed but `setup-qt-libs` standalone is not enough. 
I may make PR to `setup-qt-libs` when identify which one is missed.

ccordoba12: Ok, no problem then. It was just a minor thing.

ccordoba12 (approved):
Thanks @Czaki for your help with this!

Czaki:
Merge (to unblock the second PR)?

ccordoba12:
Sure, let's do it! I was waiting for @dalthviz, but since he's not available this week and this doesn't affect the library itself, I think it's fine to proceed.

Issue #339, opened 2022-04-23
Advise on compiling .ui files


SanPen:
Hi, 

I'm the author of [this](https://github.com/SanPen/GridCal) project that uses PySide2.

I would like to migrate completely to `qtpy` but I'm facing some issues when compiling the `.ui` files.

I've noticed that If I compile the `.ui`files with the commands provided by PySide2, then change the `.py`includes to use `qtpy` and then run the project with PySide6, some elements do not transfer well.

I think that the main issue is that I do not have a cross-qt compiler of the `.ui`files to `.py` that are fully compatible with `qtpy`.

Is there some advise on how to handle this?

CAM-Gerlach:
@dalthviz ?

dalthviz:
Hi @SanPen, quite nice project! About your question, well that's kind of tricky, I haven't used .ui files to be honest and I think there is no way at the moment to parse the .ui files into a qtpy supported version, sorry :/ (Maybe @ccordoba12 knows better).

Also, are you trying to drop generating the GUI by using .ui files and trying to move the last generated files to use qtpy then? If that is the case maybe we could try to help you checking the errors that you are seeing when running with PySide6 (maybe some of them are caused due to missing things on the qtpy side (missing imports that are not being exposed here yet or namespace errors due to a name difference between qtpy and PySide6, etc).

Just in case, on `qtpy.uic` we expose some methods to parse .ui files but at the end those will call either the PyQt or the PySide version methods.

Anyhow If there is something we can help or you see that we can add in qtpy fix things appearing in your migration process let us know!

dalthviz:
Just in case, maybe this could be interesting: https://github.com/hmaarrfk/qtpy-tools/

zkovari:
I also use a lot of `ui` files for PyQt5/6, but I always recompile them on CI without pushing the generated code to Git. So that also means that switching the underlying binding (pyqt5/6, pyside) is easy because the UI files are already binding-independent. 

I've recently added a little library that compiles multiple UI files recursively and generates code compatible with pyqt5/6 or pyside2/6 - depending on what you have on the path. It's only me using it atm, so obviously it needs some work, doc, and better tests. But it's actually pretty simple: https://github.com/plotlyst/qt-uic/blob/main/qtuic/__init__.py
The usage would be:
```
from qtuic import compile_dir

compile_dir('ui', 'qtuic/generated', recursive=True)
```
[...]

Issue #265, opened 2021-10-28
Investigate and (hopefully) resolve segfaults and other issues in UIC tests


CAM-Gerlach:
Originally discussed in PR #262 .

At present, a number of the UIC tests are skipped for various reasons, which might need to be investigated. Most prominently, the tests segfault on our Linux CIs (unconfirmed if this also occurs locally) under all Qt5 and Qt6 bindings. Furthermore, `test_load_ui_type` fails on Pyside2/Pyside6 on most platforms, due to what looks like missing Pyside2uic, and oddly, `test_load_custom_ui_auto` fails on (aside from Linux CIs) one very specific platform: Python 3.9 + macOS + pip with a error message that seems like it might be solvable ([here's the check run](https://github.com/spyder-ide/qtpy/actions/runs/1383343631). Perhaps at least some of these issues can be investigated and eliminated, preferably by someone wiser than I about Qt.

PR #440, opened 2023-07-06
PR: Replace custom implementation with loadUiType from PySide6

### LINKED ISSUE #439: qtpy.uic.loadUiType failing for PySide6
Hi, while trying to use vitables after an upgrade to PySide6 I noticed that `qtpy.uic.loadUiType` is failing:
```
  File "/home/jschmidt/mambaforge/envs/CPG_stable/lib/python3.11/site-packages/vitables/vtwidgets/renamedlg.py", line 63, in <module>
    loadUiType(os.path.join(os.path.dirname(__file__), 'rename_dlg.ui'))[0]
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jschmidt/mambaforge/envs/CPG_stable/lib/python3.11/site-packages/qtpy/uic.py", line 278, in loadUiType
    compileUi(fd, code_stream, indent=0, from_imports=from_imports)
    ^^^^^^^^^
NameError: name 'compileUi' is not defined. Did you mean: 'compile'?
```
This seems to be a bug in qtpy as `compileUi` is indeed only imported in the PySide2 branch of the imports. Or shouldn't `loadUiType` be used with PySide6?

I'm using (installed from conda-forge):
Python 3.11.4
qtpy 2.3.1
PySide6 6.5.1 
vitables 3.0.2

JaRoSchm:
This replaces the custom implementation of qtpy.uic.loadUiType with the one from PySide6. Fixes #439.

### DIFF
--- a/qtpy/tests/test_uic.py
+++ b/qtpy/tests/test_uic.py
@@ -4,9 +4,11 @@ import sys
 import warnings
 
 import pytest
+from packaging.version import parse
 
-from qtpy import PYSIDE6, PYSIDE2, QtWidgets
+from qtpy import PYSIDE6, PYSIDE2, QtWidgets, PYSIDE_VERSION
 from qtpy.QtWidgets import QComboBox
+from qtpy.tests.utils import using_conda
 
 if PYSIDE2:
     pytest.importorskip("pyside2uic", reason="pyside2uic not installed")
@@ -56,8 +58,9 @@ def test_load_ui(qtbot):
 
 
 @pytest.mark.skipif(
-    PYSIDE2 or PYSIDE6,
-    reason="PySide2uic not consistently installed across platforms/versions")
+    PYSIDE6 and using_conda() and parse(PYSIDE_VERSION) < parse("6.5")
+    and (sys.platform in ("darwin", "linux")),
+    reason="pyside6-uic command not contained in all conda-forge packages.")
 def test_load_ui_type(qtbot):
     """
     Make sure that the patched loadUiType function behaves as expected with a
--- a/qtpy/uic.py
+++ b/qtpy/uic.py
@@ -81,6 +81,7 @@ else:
     if PYSIDE6:
         from PySide6.QtCore import QMetaObject
         from PySide6.QtUiTools import QUiLoader
+        from PySide6.QtUiTools import loadUiType
     elif PYSIDE2:
         from PySide2.QtCore import QMetaObject
         from PySide2.QtUiTools import QUiLoader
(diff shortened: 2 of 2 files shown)


dalthviz (commented):
Thank you @JaRoSchm for working on the fix! Could you also as part of this PR try to remove the skip for the test related with `loadUiType` at https://github.com/spyder-ide/qtpy/blob/b303b625f461e29bad7d9ad96598aa07a4a00159/qtpy/tests/test_uic.py#L58-L60

Other than that this LGTM :+1: Thank you again!

JaRoSchm:
Okay, looks like the pyside6-uic command which is necessary for PySide6.QtUiTools.loadUiType (see https://doc.qt.io/qtforpython-6/PySide6/QtUiTools/loadUiType.html) is not contained in the conda-forge version of PySide6 6.4.3 for Linux and macOS. Locally I use PySide6 6.5.1 on Ubuntu also installed from conda-forge and here it is present. Later I will try to find out more systematically in which versions it is present. This is probably what is ment by "PySide2uic not consistently installed across platforms/versions". This could mean that we should still skip the test (and sadly no vitables for my setup right now without larger changes on their side).

JaRoSchm:
For context there was some discussion about this in https://github.com/spyder-ide/qtpy/issues/339, https://github.com/spyder-ide/qtpy/issues/265, https://github.com/spyder-ide/qtpy/pull/262.

dalthviz:
Thank you for checking the skip @JaRoSchm ! Could you update the test skip then to follow up your findings and update the reason string to take into account the need for `pyside6-uic` command?

If that ends being up to complicated or you end up needing help let me know and I will try to help checking more in detail the skip definition :+1:

JaRoSchm:
I asked in https://github.com/conda-forge/pyside2-feedstock/issues/193 where pyside6-uic should be available and it turned out that it is only packaged for Linux and macOS in version >= 6.5. QtPy uses 6.4 for its tests. 

I tried to include this into the skip condition but somehow this only fixed the Python 3.7 tests. I don't get why the condition isn't true under Python 3.11. Could you please help me there?

qtpy/tests/test_uic.py
```diff
@@ -56,8 +58,9 @@ def test_load_ui(qtbot):
 
 
 @pytest.mark.skipif(
-    PYSIDE2 or PYSIDE6,
-    reason="PySide2uic not consistently installed across platforms/versions")
+    PYSIDE6 and using_conda() and parse(PYSIDE_VERSION) <= parse("6.4")
```
dalthviz, line 61: I think the tests are failing with Python 3.11 since there we use PySide6 `6.4.3` which is not <= `6.4`. On the other hand with Python 3.7 we test using PySide6 `6.4.0` so that skips the test. So something like this should be enough for the test to pass:

```suggestion
    PYSIDE6 and using_conda() and parse(PYSIDE_VERSION) <= parse("6.4.3")
```

dalthviz (commented):
Thank you again @JaRoSchm for all the effort here! Left a suggestion for the tests which hopefully will make the checks green🤞🏼

qtpy/tests/test_uic.py
JaRoSchm: Ah thanks, stupid mistake. I tried a more general solution which is compatible with possible future releases of 6.4.*.

dalthviz (approved):
Thank you @JaRoSchm !

PR #444, opened 2023-08-14
PR: Restore `Qt.ItemFlags` access as `Qt.ItemFlag` alias (PyQt6)

dalthviz:
Part of #442

PR #449, opened 2023-08-22
PR: Improve enum to flags aliasing for PyQt6 and PySide6 > 6.3 and `QFileDialog` static methods kwarg compatibility

dalthviz:
* Follow-up #448 and #444 
* Fix wrapper functions for `QFileDialog` static methods (used for compatibility with the `dir` vs `directory` kwarg specification)

Need for changes discovered at https://github.com/napari/napari/pull/6164#issuecomment-1688468052 and https://github.com/napari/napari/pull/6164#issuecomment-1689044857


ccordoba12 (commented):
Thanks for your work @dalthviz! It looks good to me, but I have one quick question:

Has this been tested in Napari and works as expected now? I say it because it seems your previous PR was not enough for what they need.

Issue #447, opened 2023-08-21
Type hints not working for `qtpy` specific imports


wkrasnicki:
Hi,

I want to start an incremental migration from PyQt5 using your amazing project. However, I immediately hit an obstacle, when renaming my imports. The problem is with the abstracted `Signal`, `Slot` etc. classes. Their type hints are not picked up by VSCode's intellisense, which is a bit of a deal breaker for me :/

I looked into your `QtCore` file and it seems that this is caused by exporting Signal as both `pyqtSignal` from PyQt5 and `Unknown` from other packages. When I comment out other imports, I get the correct type, but I'm not familiar enough with Python typing hacks to figure out how to solve it. Is it even possible? I would very much like this feature and could submit a PR if you have some advice what to try.

dalthviz:
Hi @wkrasnicki thank you for the feedback! Not totally sure how to proceed here (my knowledge about Python typings checking is quite limited to be honest) but which are the imports you are commenting out to enable the type checking to work? Maybe we just need to add a `if not TYPE_CHECKING` validation to such lines to ensure the typing works but also still have the actual imports?

Any other info is greatly appreciated! Let us know!

wkrasnicki:
Hello, I have produced a minimal example of the issue. It requires creating 3 files and assumes only PyQt5 is installed.

```
# env.py
PYQT5 = 1 == 1
PYQT6 = 1 == 2

# qtcore.py
from .env import PYQT6, PYQT5

if PYQT5:
    from PyQt5.QtCore import pyqtSignal as Signal
elif PYQT6:
    from PyQt6.QtCore import pyqtSignal as Signal

# main.py
from .qtcore import Signal

Signal()
```

When launched in VSCode with basic type checking, I expect this to show no red lines from type checking. There is a minor difference in the actual `qtpy`, where the variables from `env.py` are of type `Literal[False]` and `Literal[True]` respectively, but changing them to bool did not help, so here I am also using bools. I'm trying out different things, but no progress so far.

dalthviz:
Thank you for the info @wkrasnicki ! I checked and I'm not seeing any red lines 🤔 Just in case, in your actual usage of QtPy do you have the config for mypy and QtPy set on VSCode? For more info you can check: https://github.com/spyder-ide/qtpy/blob/master/README.md#type-checker-integration (more specifically the link to the Pyright config comment).

Let us now if the info above helps!

dalthviz:
[...]

PR #450, opened 2023-08-24
PR: Update Docs with information on Pyright/Pylance integration and add CLI subcommand to help generate the related config

wkrasnicki:
My proposition of updated README and CLI that explains how to update the pyright configuration to eliminate type checking errors, especially in VSCode.

Solution based on this [comment](https://github.com/spyder-ide/qtpy/issues/352#issuecomment-1170684412) and follows the discussion from my [issue](https://github.com/spyder-ide/qtpy/issues/447).

NOTE:
I couldn't get the pre-commit hooks to run. It seems that some of them are outdated and raise this [error](https://github.com/PyCQA/isort/issues/2077) that has been fixed in newer versions. After updating, however, the yamllint check was throwing errors instead, so I left it as is.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,6 +104,11 @@ conda install qtpy
 
 ### Type checker integration
 
+Type checkers have no knowledge of installed packages, so these tools require
+additional configuration.
+
+#### Mypy
+
 A Command Line Interface (CLI) is offered to help with usage of QtPy.
 Presently, its only feature is to generate command line arguments for Mypy
 that will enable it to process the QtPy source files with the same API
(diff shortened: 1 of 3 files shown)


dalthviz (commented):
Great work here @wkrasnicki ! Thank you so much for taking the time to not only update the README but also extend the CLI support to get the configs! I think the only missing element here is to add a test at https://github.com/spyder-ide/qtpy/blob/master/qtpy/tests/test_cli.py for the new CLI command but other than that this LGTM :+1:

dalthviz:
> NOTE:
I couldn't get the pre-commit hooks to run. It seems that some of them are outdated and raise this https://github.com/PyCQA/isort/issues/2077 that has been fixed in newer versions. After updating, however, the yamllint check was throwing errors instead, so I left it as is.

That makes sense, we planned to use `pre-commit` here at some point (adding also a CI check for that) but that is still a pending task. Here the related issue: https://github.com/spyder-ide/qtpy/issues/345

I think we added the `.pre-commit-config.yaml` file while merging some other things 😅 but maybe could be worthy to remove it for the moment since it is throwing errors when used 🤔

What do you think @ccordoba12 @CAM-Gerlach ?

wkrasnicki:
Ah, the power of tests. It actually caught a bug with boolean values not being lowercase in TOML 😅

wkrasnicki:
Btw I have black formatter configured in my IDE to format on save and it looks like it changed the formatting a bit. Sorry for that, would you like me to revert these changes?

dalthviz:
> Btw I have black formatter configured in my IDE to format on save and it looks like it changed the formatting a bit. Sorry for that, would you like me to revert these changes?

I did a quick check and seem small enough, so no problem with just leaving them :+1:

dalthviz (approved):
Thank you for all the work here @wkrasnicki !

ccordoba12:
> What do you think @ccordoba12 @CAM-Gerlach ?

I think that work should be finished since more and more people are contributing to this project. Formatting the code with Black and then adding pre-commit should be enough.

wkrasnicki:
And thank you for all the support and prompt responses @dalthviz

Issue #458, opened 2023-09-08
Typo causes failed QWebEngineScript with PySide6


cafeclimber:
As the title says, `QWebEngineScrip` should be `QWebEngineScript`

https://github.com/spyder-ide/qtpy/blob/540fe7e560168e3ba27c17f6e423ac5f0b27620c/qtpy/QtWebEngineWidgets.py#L67

dalthviz:
Hi @cafeclimber thank you for the feedback! I think this has been recently fixed by PR #455 . The fix will be available in the next QtPy release (2.4.1). I will link this issue with #455 and mark it as part of the v2.4.1 milestone for future reference :+1:

tghw:
This is a complete breaker for Pyside6 and has been over a month since filed. Any chance we could get a release to fix this issue sooner?

dalthviz:
Hi @tghw thank you for the feedback! I think that seems reasonble, we will try to do a bug fix release (2.4.1) soon  :+1:

tghw:
@dalthviz Thanks!

For anyone who comes across this in the meantime, monkey patching the `PySide6.QtWebEngineCore` module with the typo'd name does work, but it's an ugly fix.

PR #460, opened 2023-10-13
Improve add_action implementation

ramses44:
Simplify code, reduce cyclomatic complexity from level C to A of add_action()

### DIFF
--- a/qtpy/_utils.py
+++ b/qtpy/_utils.py
@@ -81,6 +81,13 @@ def add_action(self, *args, old_add_action):
     shortcut: QKeySequence | QKeySequence.StandardKey | str | int
     receiver: QObject
     member: bytes
+
+    # if args and isinstance(args[0], QIcon):
+    if any(map(lambda arg: isinstance(arg, QIcon), args[:1])):  # Better to use previous line instead of this
+        icon, *args = args
+    else:
+        icon = QIcon()
+
     if all(
         isinstance(arg, t)
         for arg, t in zip(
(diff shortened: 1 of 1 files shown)


ramses44:
pre-commit.ci autofix

dalthviz (commented):
Thank you @ramses44 for checking how to improve this! Left a question regarding some comments that where left in the changes. Also, maybe adding more test to this util could be worthy? Besides that this looks good to me :+1: Thank you again for giving this a check!

Also, what do you think about this @ccordoba12 ? and sorry for the out of the blue ping @StSav012 but do you have any comment on these changes? Probably you have a better understanding of the shortcomings that could come when changing this util function :)

StSav012 (changes requested):
The code, as it is, contains at least two logical errors. Moreover, the code and the tests should be improved to accommodate the case of a `Qt.Key` instance as the `shortcut` argument.

dalthviz:
> Should this be a blocker for v2.4.1 given the breakage that's evidently causing per https://github.com/spyder-ide/qtpy/issues/458 ?

I though that maybe this was a kind of simple improvement but now I see that this would need way more work to accomodate things with the new findings (thanks @StSav012 for giving this a check!). So not a blocker :+1:

ccordoba12:
> Also, what do you think about this @ccordoba12 ?

Since (according to the description) this is refactoring, @StSav012 knows this code quite well and his review shows that it has several problems, and @StSav012's PR #461 conflicts with this one, I think we should close it in favor of that one.

PR #473, opened 2024-01-16
PR: Make `QApplication.exec_` a simple alias for `QApplication.exec` (Qt6)

rear1019:
See commit message for details. The change is required for Spyder to start with PyQt6/PySide6. Instead of changing `qtpy`, we also could opt for removing the monkey patch of `QApplication` in Spyder: Spyder seems to work without it (However, mind that I haven’t checked beyond “Spyder starts”).

[1] See https://github.com/spyder-ide/spyder/blob/e624b704959c965ce5a83991f9e374a7759dc6cc/spyder/app/utils.py#L302-L314

### DIFF
--- a/qtpy/QtWidgets.py
+++ b/qtpy/QtWidgets.py
@@ -79,11 +79,7 @@ elif PYQT6:
         *args,
         **kwargs,
     )
-    QApplication.exec_ = lambda *args, **kwargs: possibly_static_exec(
-        QApplication,
-        *args,
-        **kwargs,
-    )
+    QApplication.exec_ = QApplication.exec
     QDialog.exec_ = lambda self, *args, **kwargs: self.exec(*args, **kwargs)
     QMenu.exec_ = lambda *args, **kwargs: possibly_static_exec(
         QMenu,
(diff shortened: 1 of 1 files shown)


dalthviz (commented):
Hi @rear1019 sorry for such a late response and thank you for your help here! I think the `exec_` handling via the helper was added at https://github.com/spyder-ide/qtpy/pull/422 but if I understand correctly then seems like its use is not needed at least for `QApplication` and it can cause issues for subclasses, right?

Could it be possible to add a test to check for the `QApplication` subclassing and  `exec/exec_` calling case please?

Also, just in case what do you think @ccordoba12 @StSav012 ?

Thanks again for your work here!

StSav012:
@rear1019, thank you for bringing the issue up. I haven't dag deeper than the following paragraph.

Well, the `exec` method is actually static [[`QCoreApplication.exec`](https://doc.qt.io/qt-6/qcoreapplication.html#exec), [`QGuiApplication.exec`](https://doc.qt.io/qt-6/qguiapplication.html#exec), [`QApplication.exec`](https://doc.qt.io/qt-6/qapplication.html#exec)]. I'm genuinely surprised that no one calls it this way. I don't.

However, in `PySide6`, `QCoreApplication.exec_` (with the underscore) is _not_ static. [Report it](https://bugreports.qt.io/) if you wish to. That's yet another incompatibility between the flavors. I don't have time to add the fix to `QtPy` until May, probably.

Therefore, I thought that as `QCoreApplication.exec` _are_ static, the `possibly_static_exec` is semantically appropriate there.

You're totally right that as the calls are always to the static method there, the `possibly_static_exec` is excessive there. The PR looks good, although I'd add comments into the code to describe the inconsistency between the `exec_` implementations for different `QApplication`, `QDialog`, and `QMenu`.

Finally, please, change `QGuiApplication.exec` and `QCoreApplication.exec` similarly.

ccordoba12:
@dalthviz, do you want to address this one for 2.4.2 (according to @rear1019's comments) or leave it for later?

PR #475, opened 2024-03-01
PR: Add `QtCore.Qt.MouseButton.MidButton` alias for Qt6

PierreRaybaut:
If QtCore.Qt.MidButton works with QtPy and Qt6, then QtCore.Qt.MouseButton.MidButton is expected to work as well.

### DIFF
--- a/qtpy/QtCore.py
+++ b/qtpy/QtCore.py
@@ -87,6 +87,7 @@ elif PYQT6:
 
     # Alias for MiddleButton removed in PyQt6 but available in PyQt5, PySide2 and PySide6
     Qt.MidButton = Qt.MiddleButton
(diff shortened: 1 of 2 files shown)


dalthviz (commented):
Hi @PierreRaybaut ! Sorry for the late response and thank you for the fix! Could you also add the alias to the test at https://github.com/spyder-ide/qtpy/blob/3238de7a3e038daeb585c1a76fd9a0c4baf22f11/qtpy/tests/test_qtcore.py#L172-L190

?

Other than that this LGTM :+1: Thanks again!

ccordoba12:
@dalthviz, please push to this PR the fix you mentioned so we can include it in 2.4.2.

ccordoba12 (approved):
Thanks @PierreRaybaut!

PR #478, opened 2024-04-13
Add `QT_VERSION` environment variable

### LINKED ISSUE #476: Allow to specify Qt Version via Environemt Variable
In some circumstances it may be useful to specify the Qt version `qtpy` should use without without specifying which implementation to use. Currently it is only possible leave it up to `qtpy` to decide or to exactly specify the API using the `QT_API` environment variable.

In my particular use case I would need this feature in an application using `QtPdf` (which is only available in Qt6) but am indifferent whether `PyQt6` or `PySide6` should be used. I am aware that it is possible to set `QT_API=pyqt6` and `pyqt` will fall back to `PySide6` when `PyQt6` is not available but I would consider this a bit unsatisfactory since it will trigger a warning for an unproblematic case, the actual intention does not become clear when looking at the code and it does not work when only `Qt5` should be used instead.

This feature could be realized by allowing to set `QT_API` to a Qt version , e.g. `QT_API=Qt6` or by allowing to set `QT_API` to a list of values, e.g. `QT_API=pyqt6,pyside6`.

Bzero:
This PR adds the option to specify the Qt version qtpy will use via the `QT_VERSION` environment variable and should resolve #476.

If the `QT_VERSTION` but not the `QT_API` environment variable is set, the `API` is selected based on `QT_VERSTION` (and follows the usual order `PyQt5` -> `PySide2` -> `PyQt6` -> `PySide6` if a binding is not installed). If both environment variables are set, `QT_API` takes precedence. If the version cannot be satisfied or the environment variables are incompatible warnings are triggered.

Please have an extra eye on the test when reviewing as I am not perfectly sure it will behave correctly in the CI environment.

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -140,9 +140,13 @@ class QtModuleNotInstalledError(QtModuleNotFoundError):
 # Qt API environment variable name
 QT_API = "QT_API"
 
+# Qt VERSION environment variable name
+QT_VERSION_ENV = "QT_VERSION"
+
 # Names of the expected PyQt5 api
 PYQT5_API = ["pyqt5"]
 
+# Names of the expected PyQt6 api
 PYQT6_API = ["pyqt6"]
 
 # Names of the expected PySide2 api
@@ -150,6 +154,12 @@ PYSIDE2_API = ["pyside2"]
 
 # Names of the expected PySide6 api
 PYSIDE6_API = ["pyside6"]
+
+# Names of the expected Qt5 version name
+QT5_VERSIONS = ["qt5"]
+
+# Names of the expected Qt6 version name
+QT6_VERSIONS = ["qt6"]
 
 # Minimum supported versions of Qt and the bindings
 QT5_VERSION_MIN = PYQT5_VERSION_MIN = "5.9.0"
@@ -160,8 +170,9 @@ QT_VERSION_MIN = QT5_VERSION_MIN
 PYQT_VERSION_MIN = PYQT5_VERSION_MIN
 PYSIDE_VERSION_MIN = PYSIDE2_VERSION_MIN
 
-# Detecting if a binding was specified by the user
+# Detecting if a binding or version was specified by the user
 binding_specified = QT_API in os.environ
+version_specified = QT_VERSION_ENV in os.environ
 
 API_NAMES = {
     "pyqt5": "PyQt5",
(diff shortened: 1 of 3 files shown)


qtpy/tests/test_main.py
```diff
@@ -140,3 +168,16 @@ def test_qt_api_environ(api):
...
+    """
+    If QT_VERSION is specified, check that one of the correct Qt wrappers was used
+    """
+
+    QT_VERSION = os.environ.get("QT_VERSION", "").lower()
+
+    if QT_VERSION == "qt5":
+        assert_qt5()
+    elif QT_VERSION == "qt6":
+        assert_qt6()
```
dalthviz, line 183: Not totally sure but, I think we are not setting the `QT_VERSION` env var over the CI so for the moment this test is passing without doing any assert?

dalthviz (commented):
Thank you @Bzero for your work here! Checked the CI failures and I think those were unrelated to the work being done here (trigger a new run and seems like checks are passing now). Left some initial comments but from a quick review I would say the changes make sense :+1:

qtpy/__init__.py
```diff
@@ -173,10 +184,42 @@ def __init__(self, *, missing_package=None, **superclass_kwargs):
...
+
+
+# Warn if the QT_API and the QT_VERSION environment variables are incompatible
+if (
+    API_NAMES[initial_api] not in VERSION_NAMES[requested_version]
+    and version_specified
+    and binding_specified
+):
+    warnings.warn(
+        f"Specified QT_API={initial_api!r} and specified QT_Version={requested_version!r} are incompatible, QT_API will take precedence.",
```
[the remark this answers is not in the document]
Bzero: Thanks for spotting this error!

qtpy/tests/test_main.py
Bzero: Yes, I don't think `QT_VERSION` is set in CI, ideally both possibilities should be checked. Do you happen to know how this works for the similar test [`test_qt_api`](https://github.com/Bzero/qtpy/blob/b8c9d20e508b66d269a250ad22f9b7c9c64ca139/qtpy/tests/test_main.py#L97)?

dalthviz: I think currently the `test_qt_api` is going through the `else` branch 😅 Checking the script that does the test setup ([test.sh](https://github.com/Bzero/qtpy/blob/b8c9d20e508b66d269a250ad22f9b7c9c64ca139/.github/workflows/test.sh)) I'm not seeing us setting up `QT_API` 🤔 Maybe we should try to check a way to setup `QT_API` (and `QT_VERSION`) for the tests (I think there is a `USE_QT_API` that can be set for test but I think we are not using it here either). What do you think @ccordoba12 ?

ccordoba12: I also think that's a good idea but I don't understand very well how to set up those env vars in our CIs.

PR #483, opened 2024-04-25
PR: Fix coverage errors and macOS jobs

dalthviz:
See https://github.com/coverallsapp/github-action/issues/205 and https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories


ccordoba12 (approved):
Thanks @dalthviz!

PR #491, opened 2024-06-07
[WIP] Adding Scorecard.yml Workflow and Scorecard Badge GOSST

PatriceJada:


dalthviz:
@ccordoba12 is this something you would like to have in QtPy? If I remember correctly I think I have seen similar PRs for this badge in other spyder-ide GitHub org repos. I think at somepoint I checked and seemed like this is something related with some sort of NumFOCUS iniciative/SDG, right? Let me know if this should get merged or closed!

Issue #494, opened 2024-10-11
Crash when importing QtCore with PySide 6.8


maflAT:
With PySide 6.8.0 installed, qtpy will crash without raising an Exception when importing QtCore from qtpy.
The crash happens in "./qtpy/QtCore.py", line 128: `from PySide6.QtCore import *`.

ccordoba12:
Hey @maflAT, thanks for reporting. This seems similar to issue #480, in the sense that PySide 6.8 seems to have broken star imports again (i.e. imports of the form from `PySide6.QtCore import *`).

Since Qtpy is simply a wrapper around PyQt and PySide, there's nothing we can do about it. Instead, please report this problem in the [Qt bug tracker](https://bugreports.qt.io/secure/Dashboard.jspa) and kindly ask them to run our test suite before releasing a new minor version to avoid this problem in the future.

hmaarrfk:
xref: https://bugreports.qt.io/browse/PYSIDE-2888

owillebo:
Note that the crash is a CPython crash, i.e. an access violation (exit code 0xC0000005) on windows, i.e.;

`import qtpy.QtCore`

crashes with 0xC0000005 and;

`import PySide6.QtCore`

exits normally with exit code 0.

See attached test.bat file.

This bat file outputs;

```
Python 3.11.9
packaging==24.1
PySide6==6.8.0
PySide6_Addons==6.8.0
PySide6_Essentials==6.8.0
QtPy==2.4.1
shiboken6==6.8.0
QtPy exit code: -1073741819
PySide6 exit code: 0
```

[test.zip](https://github.com/user-attachments/files/17351002/test.zip)

hmaarrfk:
My guess is that the 6.8.0.1 from pypi resolved things they released it today.

On conda-forge we also released a backport though we called it `6.8.0 build 1`
https://github.com/conda-forge/pyside2-feedstock/pull/249

Upstream seems to have added a test for star imports too.

hmaarrfk:
In the fact the bug report I originally opened refers to 6.8.0.1 as having the fix:
https://bugreports.qt.io/browse/PYSIDE-2888

PR #495, opened 2024-10-11
PR: Help users understand why PySide 6.8.0 gives a segfault

hmaarrfk:
xref: https://github.com/spyder-ide/qtpy/issues/494

I'm not sure if you like to add things like this, but it is pretty hard to track this bug down.

### DIFF
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -281,8 +281,8 @@ if API in PYSIDE6_API:
 
         if PYSIDE_VERSION == "6.8.0":
             print(
-                "A known critical bug in PySide6 6.8.0 will cause your application to crash "
-                "see https://github.com/spyder-ide/qtpy/issues/494",
+                "A known critical bug in PySide6 6.8.0 will cause your application to crash. "
+                "See https://github.com/spyder-ide/qtpy/issues/494",
             )
         from PySide6.QtCore import __version__ as QT_VERSION  # analysis:ignore
 
--- a/qtpy/__init__.py
+++ b/qtpy/__init__.py
@@ -278,6 +278,11 @@ if API in PYQT6_API:
 if API in PYSIDE6_API:
     try:
         from PySide6 import __version__ as PYSIDE_VERSION  # analysis:ignore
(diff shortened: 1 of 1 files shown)


qtpy/__init__.py
```diff
@@ -278,6 +278,12 @@ def __init__(self, *, missing_package=None, **superclass_kwargs):
 if API in PYSIDE6_API:
     try:
         from PySide6 import __version__ as PYSIDE_VERSION  # analysis:ignore
+
+        if PYSIDE_VERSION == "6.8.0":
+            print(
+                "A known critical bug in PySide6 6.8.0 will cause your application to crash "
+                "see https://github.com/spyder-ide/qtpy/issues/494",
```
ccordoba12, line 285:
```suggestion
                "A known critical bug in PySide6 6.8.0 will cause your application to crash. "
                "See https://github.com/spyder-ide/qtpy/issues/494",
```

qtpy/__init__.py
```diff
@@ -278,6 +278,12 @@ def __init__(self, *, missing_package=None, **superclass_kwargs):
...
+            )
```
ccordoba12, line 286: Shouldn't we add a `return` below this line to avoid the segfault?

ccordoba12 (commented):
Thanks @hmaarrfk for your help with this!

qtpy/__init__.py
ccordoba12: Actually, that's not a good idea because it'll be generated in another module. So, ignore my comment.

hmaarrfk: Yeah. I mean. I’m not trying to be fancy here. Just a bit of damage control.

hmaarrfk: We shouldn’t overthink these things

ccordoba12: Agreed. I'll release 2.4.2 with your fix during the weekend or on Monday at most.

ccordoba12 (approved):
Thanks @hmaarrfk!

PR #506, opened 2024-12-07
PR: Ignore PySide6 `QSqlDatabase.exec` `DeprecationWarning`

juliangilbey:
When running the package tests with PySide 6, the tests fail because PySide 6 gives a DeprecationWarning, and the pytest.ini settings turn this into an error.  This PR ignores that specific error.

I tried using a `warnings.catch_warnings()` block in `qtpy/QtSql.py` around the deprecated functions, but pytest seems to ignore this.  It might be worth adding such a block anyway.

### DIFF
--- a/pytest.ini
+++ b/pytest.ini
@@ -3,6 +3,7 @@ addopts = --durations=10 -v -r a --color=yes --code-highlight=yes --strict-confi
 empty_parameter_set_mark = fail_at_collect
 filterwarnings =
     error
+    ignore:.*QSqlDatabase\.exec\(:DeprecationWarning
 log_auto_indent = True
 log_level = INFO
 minversion = 6.0


dalthviz:
Hi @juliangilbey thanks for the PR! So my guess is that doing something like:

```python

def prevent_warn_exec(self, *args, **kwargs):
    with warnings.catch_warnings():
        warnings.simplefilter("ignore")
        self.exec(*args, **kwargs)

QSqlDatabase.exec_ = QSqlDatabase.exec = prevent_warn_exec

```

Didn't work, right? If that is the case then seems reasonable to just ignore the warning. Also what do you think @ccordoba12 ?

dalthviz:
Note: More info on the deprecation at https://doc.qt.io/qt-6/qsqldatabase-obsolete.html#exec

ccordoba12:
> Also what do you think @ccordoba12 ?

I think we shouldn't ignore the warning with the approach you proposed so that users actually see it and learn that calling `exec` is deprecated.

So, I'm ok with @juliangilbey's simple approach of ignoring the warning in our tests so they pass.

ccordoba12 (approved):
Thanks @juliangilbey!

PR #507, opened 2024-12-12
PR: Ensure `QMenu` and `QToolBar` `isinstance` checks succeed and remove unneeded wrapper classes 

### LINKED ISSUE #502: Hijacked PyQt types are not returned from native Qt APIs
QtPy 2.4.2 has integrated this PR: https://github.com/spyder-ide/qtpy/pull/461, which swaps `QToolbar` and `QAction` types with custom subclasses. This introduced a new problem that instances returned from Qt no longer pass instance checks.

E.g.:

```python
from qtpy import QtWidgets

app = QtWidgets.QApplication([])
window = QtWidgets.QMainWindow()
new_toolbar = window.addToolBar('Toolbar title')
assert isinstance(new_toolbar, QtWidgets.QToolBar)
```

This will pass in QtPy 2.4.1 and will fail with QtPy 2.4.2. It now requires to check against true PyQt classes, e.g. ` assert isinstance(new_toolbar, PyQt5.QtWidgets.QToolBar)`

By extension, returned instances will have different behavior from the manually created ones

### LINKED ISSUE #505: Breaking change related to QMenu in 2.4.2
Prior to QtPy 2.4.2, the following:

```python
from qtpy.QtWidgets import QApplication, QMenu

app = QApplication([])

menu = QMenu()
menu.addMenu('test')

submenu = menu.children()[1]

print(isinstance(submenu, QMenu))
```

would return ``True``, but now returns ``False`` because the class for menus added using ``addMenu`` is the native QMenu class whereas QtPy now exposes a wrapped class. Fixing this might be a matter of overriding ``addMenu``?

dalthviz:
Fixes #502
Fixes #505

### DIFF
--- a/qtpy/QtGui.py
+++ b/qtpy/QtGui.py
@@ -227,6 +227,7 @@ if PYQT5 or PYSIDE2:
 
     # Follow similar approach for `QDropEvent` and child classes
     QDropEvent.position = lambda self: self.posF()
+
 if PYQT6 or PYSIDE6:
     # Part of the fix for https://github.com/spyder-ide/qtpy/issues/394
     for _class in (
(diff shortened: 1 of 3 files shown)


StSav012:
My bad. I haven't thought of such a consequence. The changes you've proposed are the only way, then. Apparently, I've removed the test case that led me to the inheritance of the classes. So, I can't review it to see if it was erroneous. As the tests pass, it should be OK.

ccordoba12 (approved):

ccordoba12:
@dalthviz, given the severity of this problem, I think we should release 2.4.3 asap.

What do you think?

PR #518, opened 2025-10-06
PR: Add `QtStateMachine` import paths for all bindings/versions where it is available

krokosik:
The `QtStateMachine` import was defined only for `Pyside6` despite the fact that it exists in other bindings as well. This PR adds proper import paths.

### DIFF
--- a/qtpy/tests/test_qtstatemachine.py
+++ b/qtpy/tests/test_qtstatemachine.py
@@ -1,14 +1,20 @@
 import pytest
 
-from qtpy import QtBindingInNewerVersionError
+from qtpy import PYQT6, PYQT_VERSION, PYSIDE6
+from qtpy.tests.utils import using_conda
 
 
+@pytest.mark.skipif(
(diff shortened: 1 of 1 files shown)


dalthviz:
Hi @krokosik thank you for your work here and sorry for the late response! Could it be possible for you to update with latest master to fix the failing tests? Also, with the changes here I think the import of `QtStateMachine` should work with any binding so could you remove the skip import line from the `QtStateMachine` test:

https://github.com/spyder-ide/qtpy/blob/281393f92baf68f4eeaa370116e30a5a6170c731/qtpy/tests/test_qtstatemachine.py#L8

If you need any help to finish this let us know!

krokosik:
@dalthviz No problem, thank you for the review :)

I've done what you asked and also wonder whether you want to mention in the docs that `qtpy` uses the Qt6 import path (in the same way you mention discrepancies in signals etc.)?

krokosik:
Awesome, thanks :)

PR #521, opened 2025-11-05
PR: Constraint pytest-qt to 4.4.0 (<4.5.0) when testing PySide2 (CI)

dalthviz:
Checking PR https://github.com/spyder-ide/qtpy/pull/518 noticed that we are missing to handle the PySide2 drop in the latest pytest-qt version available (4.5.0)


ccordoba12 (approved):
Thanks @dalthviz!

